PR #113, opened 2014-11-24
Implement string hash code in C.

sun-rui:
[SPARKR-135](https://sparkr.atlassian.net/browse/SPARKR-135). Improve wordcount performance by implement the string hash code in C. The performance improvement is about 10X.
On a 3-worker-node Spark cluster, each node has 16 threads (2.93Ghz Xeon), 48G memory:

3.2 G data  from 96 minutes to 8 minutes
320 M data from 10 minutes to 1 minute

@shivaram I haven't test this patch on Windows, could you help to test it? Thanks.

### DIFF
--- a/pkg/src/Makefile.win
+++ b/pkg/src/Makefile.win
@@ -25,7 +25,7 @@ $(MAVEN_TARGET_NAME): $(SCALA_FILES) $(RESOURCE_FILES)
 	cp -f $(MAVEN_TARGET_NAME) ../inst/$(JAR_NAME)
 
 sharelib: string_hash_code.c
-	R CMD SHLIB -o SparkR.so string_hash_code.c
+	R CMD SHLIB -o SparkR.dll string_hash_code.c
 
 clean:
 	mvn.bat clean
@@ -35,6 +35,6 @@ clean:
 	-rm sbt/sbt-launch-*.jar
 	rm -f ../inst/$(JAR_NAME) 
 	rm -f *.o
-	rm -f *.so
+	rm -f *.dll
        
 .PHONY: all clean


7c00:
This patch works well on my laptop. But when I compiled the pkg, for the first time, on a fresh installed server. An error occurred to me. That is:

> ** testing if installed package can be loaded
Error in dyn.load(file, DLLpath = DLLpath, ...) : 
  unable to load shared object '/home/todd/git/SparkR-pkg/lib/SparkR/libs/SparkR.so':
  /home/todd/git/SparkR-pkg/lib/SparkR/libs/SparkR.so: undefined symbol: IS_SCALAR
Error: loading failed
Execution halted
ERROR: loading failed

The server is running ubuntu 14.04 32bit, with `R version 3.0.2 (2013-09-25)`.

shivaram:
@sun-rui can probably give more details, but I think this is due to the older R version (3.0.2) not having the `IS_SCALAR` symbol. Could you try with a newer version say 3.1.2 ? 

Also we should try to see if we can just symbols present in earlier R versions if this is the problem.

7c00:
It works after I upgraded R to 3.1.2 !

By the way, is there anything like `make clean`? I had to `rm *.so *.o` by hand now.

sun-rui:
@7c00, you can "cd pkg/src; make clean" 

I was not aware that the IS_SCALAR macro is not supported by earlier versions. But I think this is not an urgent issue to make the code work with those versions. If we receive more user's feedback for using older versions, I can find a portable solution.

7c00:
@sun-rui Nice! :clap: 

I think it would be better to add some notes or suggestions on R version to README.md before a portable solution comes up.

shivaram:
The `DESCRIPTION` file lists the R version required by a package. Technically we only require R >= 3.0, and I think its good to be backwards compatible with a few R versions. 
One simple fix could be to just import the definition to our file from https://github.com/wch/r-source/blob/70e7151b62fd95ab5660c040073d746073b11032/src/include/Rinternals.h#L319 
Lets open a JIRA to track this.

sun-rui:
Submitted https://sparkr.atlassian.net/browse/SPARKR-166 for this issue.

PR #119, opened 2014-12-17
Ability to more flexibly define the path to a text file

dputler:
This allows a user to use tilde expansion or simply the file name if the text file is in the user working directory.


pkg/R/context.R
```diff
@@ -19,6 +19,10 @@
 #'}
 
 textFile <- function(sc, path, minSplits = NULL) {
+  # Allow the user to have a more flexible definiton of the text file path
+  path <- path.expand(path)
+  if (!(grepl("\\\\", path) | grepl("/", path)))
```
dputler: Hi Shivaram,

 

I’m fine with removing the lines of code that pre-pend the current working directory and going with the path expansion. The one difference in what you and I have done is that I ran:

 

sudo R CMD INSTALL pkg

 

from within the SparkR-pkg directory, so the SparkR package is the main R library. At this point, after a partial, unsuccessful, sync of my fork with the upstream copy, my local copy is a mess, so I can’t test whether that is causing the difference in what you and I found.

 

Should I retract this pull request, and submit a new one, or is this a change you can make?

 

Dan

shivaram: @dputler - No worries about the git client. I can fix things up during merge. One last thing I wanted to check with you was whether using `normalizePath` is better than using `path.expand`. From the documentation it looks like `normalizePath` calls `path.expand` first and then also handles things like `../` etc.

dputler: HI Shivaram,

 

normalizePath does look like a better choice for this.

 

Dan

shivaram: Cool. I'll make the switch from path.expand to normalizePath during the merge.

shivaram:
Thanks @dputler -- I merged this manually with the changes discussed.

PR #128, opened 2015-01-12
[SPARKR-165] IS_SCALAR is not present in R before 3.1

sun-rui:
Verfied with R 3.0.0. Built passed and PI example can run.


shivaram:
Thanks @sun-rui  for the fix. LGTM. I also verified by looking at R source code for `IS_SCALAR`
https://github.com/wch/r-source/blob/R-3-1-branch/src/include/Rinternals.h#L345

PR #129, opened 2015-01-13
[SPARKR-151] Add name, setName in RDD.R

lythesia:
SPARKR-151: Support ability to set, retrieve RDD name.


pkg/R/RDD.R
```diff
@@ -1844,3 +1844,53 @@ setMethod("cogroup",
             cogroup.rdd <- mapValues(groupByKey(union.rdd, numPartitions), 
                                      group.func)
           })
+
+# TODO:
+# bind to env?
```
shivaram: Is this TODO referring to saving the name in the RDD's env ?

pkg/R/RDD.R
```diff
@@ -1844,3 +1844,53 @@ setMethod("cogroup",
...
+#' name(rdd) # NULL (if not set before)
+#'}
+setGeneric("name", function(rdd) { standardGeneric("name") })
+
+#' @rdname name
+#' @aliases name,RDD
+setMethod("name",
+          signature(rdd = "RDD"),
+          function(rdd) {
+            .jcall(getJRDD(rdd), "S", "name")
```
shivaram: I can see that "S" is a shortcut for "Ljava/lang/String;", but I'd actually prefer having the full type here to be explicit.

pkg/R/RDD.R
```diff
@@ -1844,3 +1844,53 @@ setMethod("cogroup",
...
+          }
+)
+
+#' Set an RDD's name.
+#'
+#' @param rdd The RDD whose name is returned
```
shivaram: This param should be `The RDD whose name is to be set`

pkg/R/RDD.R
```diff
@@ -1844,3 +1844,53 @@ setMethod("cogroup",
...
+setGeneric("setName", function(rdd, name) { standardGeneric("setName") })
+
+#' @rdname setName
+#' @aliases setName,RDD
+setMethod("setName",
+          signature(rdd = "RDD", name = "character"),
+          function(rdd, name) {
+            .jcall(getJRDD(rdd), "Lorg/apache/spark/api/java/JavaRDD;", "setName", name)
+            rdd
+          }
```
shivaram: I think the style we use here is to put the `)` on the same line as the `}`

shivaram:
Thanks @lythesia for the change. It mostly looks good and I left some minor inline comments.

Also could you modify the title of the PR to be like `[SPARKR-151] Add name, setName in RDD.R`

pkg/R/RDD.R
lythesia: yes, like the `rdd@env$isCached`, then you may access the name without `.jcall`

pkg/R/RDD.R
lythesia: Thx, I'll fix that.

pkg/R/RDD.R
shivaram: We could do that, but I am not sure `name` will be that frequently used. I am fine with leaving it as a TODO -- You could word it as something like `TODO: Consider caching the name in the RDD's environment`

shivaram:
Thanks for the update. One minor comment: The roxygen files need to be regenerated after updating the help text. Let me know if you are having trouble with it and I can regenerate it while merging

lythesia:
Sorry just missed that, manual fixed.

shivaram:
Thanks @lythesia. Merging this

PR #130, opened 2015-01-19
[SPARKR-162] add fullOuterJoin in RDD.R

lythesia:
Support full outer join as well as { left, right } we already have.


shivaram:
Thanks @lythesia for the change -- I'll try to take a look at this later today, but I just wanted to leave a comment that we should add a test case for fullOuterJoin (similar to https://github.com/amplab-extras/SparkR-pkg/blob/master/pkg/inst/tests/test_rdd.R#L323)

cc @concretevitamin who reviewed some of the earlier join code

lythesia:
@shivaram test case added and passed

lythesia:
Sorry, I'm re-arranging files, now it should be ok.

concretevitamin:
This code looks good to me. Merging into master - thanks @lythesia.

One concern that'd be good to address in the future: the different join functions have a lot of code / logic duplication, so it'd be good to investigate ways to reuse the common part.

lythesia:
I've considered to extract out common operations on temp list in join code part as util functions, but I'm not sure whether it will introduce more overhead when generating new lists via built-in `apply` `lapply`. 
For now I choose the redundant style, maybe we can refactor them in the future.

concretevitamin:
Yes, I've filed [SPARKR-175](https://sparkr.atlassian.net/browse/SPARKR-175) for future work.

PR #131, opened 2015-01-21
[SPARKR-177] New R to Java bridge

shivaram:
This PR introduces a new R to Java bridge to be used on the driver side. The newly introduced SparkRBackend is a netty-based server that opens a socket for communicating with R. We use Java reflection to support creating objects and invoking methods on these objects. More details can be found in https://docs.google.com/document/d/1hscJX3VAp3GW2h24GvUIP1UXfcq8gmxTqml1dcDuL6I/edit

This is joint work with @sun-rui 

@concretevitamin @sun-rui -- It will be great if you can review this change for code structure, understandability and make sure I haven't missed anything that needs to be updated.


pkg/R/RDD.R
```diff
@@ -1,6 +1,46 @@
...
+    .validJobjs[[objId]] <- .validJobjs[[objId]] + 1
+  } else {
+    .validJobjs[[objId]] <- 1
+  }
+  newObj
+}
+
+# Handler for a java object that exists on the backend.
+jobj <- function(objId) {
+  if (!is.character(objId)) stop("object id must be a character")
```
sun-rui: use {} for even a single statement?

pkg/R/RDD.R
```diff
@@ -1,6 +1,46 @@
...
+  obj <- structure(new.env(parent = emptyenv()), class = "jobj")
```
sun-rui: Do we really need to create a new environment for a new jobj? could we simply use obj <- objId and class(obj)<-"jobj"?

pkg/R/RDD.R
```diff
@@ -1,6 +1,46 @@
...
+  reg.finalizer(obj, cleanup.jobj)
+  obj
+}
+
+cleanup.jobj <- function(e) {
+  objId <- e$id
+  .validJobjs[[objId]] <- .validJobjs[[objId]] - 1
+
+  if (.validJobjs[[objId]] == 0) {
+    rm(list=objId, envir = .validJobjs)
```
sun-rui: list = objId

pkg/R/RDD.R
```diff
@@ -98,48 +138,44 @@ setMethod("getJRDD", signature(rdd = "PipelinedRDD"),
             computeFunc <- function(split, part) {
               rdd@func(split, part)
             }
-            serializedFunc <- serialize("computeFunc", connection = NULL,
+            serializedFuncArr <- serialize("computeFunc", connection = NULL,
                                         ascii = TRUE)
```
sun-rui: why ascii = true here?

pkg/R/RDD.R
```diff
@@ -98,48 +138,44 @@ setMethod("getJRDD", signature(rdd = "PipelinedRDD"),
...
-            serializedFuncArr <- .jarray(serializedFunc)
 
-            packageNamesArr <- .jarray(serialize(.sparkREnv[[".packages"]],
-                                                 connection = NULL,
-                                                 ascii = TRUE))
+            packageNamesArr <- serialize(.sparkREnv[[".packages"]],
+                                         connection = NULL,
+                                         ascii = TRUE)
```
sun-rui: why ascii=true here?

pkg/R/RDD.R
```diff
@@ -98,48 +138,44 @@ setMethod("getJRDD", signature(rdd = "PipelinedRDD"),
...
+                                   depsBin,
+                                   packageNamesArr,
+                                   as.character(.sparkREnv[["libname"]]),
+                                   broadcastArr,
+                                   callJMethod(prev_jrdd, "classTag"))
             }
             # Save the serialization flag after we create a RRDD
             rdd@env$serialized <- dataSerialization
-            rdd@env$jrdd_val <- rddRef$asJavaRDD()
+            rdd@env$jrdd_val <- callJMethod(rddRef, "asJavaRDD") # rddRef$asJavaRDD()
```
sun-rui: is "asJavaRDD" really needed? seems we can use RDD directly

pkg/R/RDD.R
```diff
@@ -1286,48 +1311,44 @@ setMethod("partitionBy",
...
 
-            serializedHashFunc <- serialize(as.character(substitute(partitionFunc)),
-                                            connection = NULL,
-                                            ascii = TRUE)
-            serializedHashFuncBytes <- .jarray(serializedHashFunc)
-
-            packageNamesArr <- .jarray(serialize(.sparkREnv$.packages,
+            serializedHashFuncBytes <- serialize(as.character(substitute(partitionFunc)),
                                                  connection = NULL,
-                                                 ascii = TRUE))
```
sun-rui: ascii = true can be removed?

pkg/R/RDD.R
```diff
@@ -1286,48 +1311,44 @@ setMethod("partitionBy",
...
-            refs <- lapply(ls(.broadcastNames), function(name) {
-                           get(name, .broadcastNames) })
-            broadcastArr <- .jarray(refs,
-                                    "org/apache/spark/broadcast/Broadcast")
-            jrdd <- getJRDD(rdd)
+                                                 ascii = TRUE)
```
sun-rui: ascii=true can be removed?

pkg/R/RDD.R
```diff
@@ -1286,48 +1311,44 @@ setMethod("partitionBy",
...
 
+            packageNamesArr <- serialize(.sparkREnv$.packages,
+                                         connection = NULL,
+                                         ascii = TRUE)
```
sun-rui: ascii=true can be removed?

pkg/R/RDD.R
```diff
@@ -1286,48 +1311,44 @@ setMethod("partitionBy",
...
 
             # Create a corresponding partitioner.
-            rPartitioner <- new(J("org.apache.spark.HashPartitioner"),
-                                as.integer(numPartitions))
+            rPartitioner <- newJObject("org.apache.spark.HashPartitioner",
+                                       as.integer(numPartitions))
 
             # Call partitionBy on the obtained PairwiseRDD.
-            javaPairRDD <- pairwiseRRDD$asJavaPairRDD()$partitionBy(rPartitioner)
+            javaPairRDD <- callJMethod(pairwiseRRDD, "asJavaPairRDD")
```
sun-rui: no need to convert as JavaPairRDD?

pkg/R/sparkR.R
```diff
@@ -6,14 +6,22 @@ sparkR.onLoad <- function(libname, pkgname) {
...
   .sparkREnv$libname <- libname
   .sparkREnv$assemblyJarPath <- assemblyJarPath
-  .jinit(classpath=assemblyJarPath, parameters=paste("-Xmx", sparkMem, sep=""))
-  .jaddClassPath(yarn_conf_dir)
+}
+
+#' Stop this Spark context
+#' Also terminates the backend this R session is connected to
+sparkR.stop <- function(sparkREnv) {
+  sc <- get(".sparkRjsc", envir=.sparkREnv)
```
sun-rui: need to check if sc exists. sparkR.init() may not be called before.

pkg/R/sparkR.R
```diff
@@ -6,14 +6,22 @@ sparkR.onLoad <- function(libname, pkgname) {
...
+  callJMethod(sc, "stop")
+  callJStatic("SparkRHandler", "stopBackend")
+
+  # Also close the connection and remove it from our env
+  conn <- get(".sparkRCon", .sparkREnv)
```
sun-rui: need to check if conn exists. sparkR.init() may not be called before.

pkg/R/sparkR.R
```diff
@@ -52,6 +60,17 @@ sparkR.init <- function(
...
   }
 
+  sparkMem <- Sys.getenv("SPARK_MEM", "512m")
+  cp = .sparkREnv$assemblyJarPath
+  yarn_conf_dir <- Sys.getenv("YARN_CONF_DIR", "")
+  if (yarn_conf_dir != "") {
+    cp = paste(cp, yarn_conf_dir, sep=":")
+  }
+  launchBackend(classPath=cp, mainClass="edu.berkeley.cs.amplab.sparkr.SparkRBackend",
+                args="12345", javaOpts=paste("-Xmx", sparkMem, sep=""))
```
sun-rui: could we add a configuration allowing the user to specify a port in the case of port conflict?

pkg/R/sparkR.R
```diff
@@ -60,20 +79,20 @@ sparkR.init <- function(
...
-    sparkExecutorEnvMap$put("LD_LIBRARY_PATH", paste0("$LD_LIBRARY_PATH:",Sys.getenv("LD_LIBRARY_PATH")))
+    sparkExecutorEnvMap[["LD_LIBRARY_PATH"]] <- paste0("$LD_LIBRARY_PATH:",Sys.getenv("LD_LIBRARY_PATH"))
   }
   for (varname in names(sparkExecutorEnv)) {
-    sparkExecutorEnvMap$put(varname, sparkExecutorEnv[[varname]])
+    sparkExecutorEnvMap[[varname]] <- sparkExecutorEnv[[varname]]
   }
   
-  .jaddClassPath(sparkJars)
+  #.jaddClassPath(sparkJars)
```
sun-rui: Add sparkJars to the classpath when lanuching the backend?

pkg/R/sparkR.R
```diff
@@ -81,16 +100,22 @@ sparkR.init <- function(
 
   assign(
     ".sparkRjsc",
-    J("edu.berkeley.cs.amplab.sparkr.RRDD",
-      "createSparkContext",
+    callJStatic(
+      className="edu.berkeley.cs.amplab.sparkr.RRDD",
```
sun-rui: space?

pkg/R/sparkR.R
```diff
@@ -81,16 +100,22 @@ sparkR.init <- function(
...
+      methodName="createSparkContext",
```
sun-rui: space?

pkg/R/sparkRBackend.R
```diff
@@ -0,0 +1,85 @@
...
+}
+
+# Call a static method on a specified className
+callJStatic <- function(className, methodName, ...) {
+  invokeJava(isStatic=TRUE, className, methodName, ...)
+}
+
+# Create a new object of the specified class name
+newJObject <- function(className, ...) {
+  invokeJava(isStatic=TRUE, className, methodName="new", ...)
```
sun-rui: change the method name "new" to "<init>" (constructor name in byte code) in case there may be a method called "new" in a class?

pkg/R/sparkRBackend.R
```diff
@@ -0,0 +1,85 @@
...
+  # If this is already isn't a removeJObject call
+  if (!isRemoveMethod(isStatic, objId, methodName)) {
+    objsToRemove <- ls(.toRemoveJobjs)
+    if (length(objsToRemove) > 0) {
+      #cat("Clearing up objsToRemove\n")
+      sapply(objsToRemove,
+            function(e) {
+              removeJObject(e)
+            })
+      rm(list=objsToRemove, envir=.toRemoveJobjs)
```
sun-rui: space between =?

pkg/R/utils.R
```diff
@@ -16,51 +16,41 @@ convertJListToRList <- function(jList, flatten, logicalUpperBound = NULL, serial
...
-
-               # For serialized datasets, `obj` (and `rRaw`) here corresponds to
-               # one whole partition dense-packed together. We deserialize the
-               # whole partition first, then cap the number of elements to be returned.
-
-               # TODO: is it possible to distinguish element boundary so that we can
-               # unserialize only what we need?
-               if (!is.null(logicalUpperBound)) {
+             if (inherits(obj, "jobj")) {
+               if (grepl("scala.Tuple2", obj$id)) {
```
sun-rui: Could we add a RPC method for instanceOf?  using grepl seems not work for subclasses

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SerializeJavaR.scala
```diff
@@ -0,0 +1,296 @@
...
+
+  def readDouble(in: DataInputStream) = {
+    in.readDouble()
+  }
+
+  def readString(in: DataInputStream) = {
+    val len = in.readInt()
+    val asciiBytes = new Array[Byte](len)
+    in.read(asciiBytes, 0, len)
+    val str = new String(asciiBytes.dropRight(1).map(_.toChar))
```
sun-rui: add an assert for the ending '\0'?

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
```diff
@@ -0,0 +1,205 @@
...
+    // First bit is isStatic
+    val isStatic = readBoolean(dis)
+    val objId = readString(dis)
+    val methodName = readString(dis)
+    val numArgs = readInt(dis)
+
+    if (objId == "SparkRHandler") {
+      methodName match {
+        case "stopBackend" => {
+          // dos.write(0)
```
sun-rui: remove this line

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
```diff
@@ -0,0 +1,205 @@
...
+  def getObject(id: String): Object = {
+    objMap(id)
+  }
+
+  def get(id: String): Option[Object] = {
+    objMap.get(id)
+  }
+
+  def put(obj: Object): String = {
+    val objId = obj.getClass().getName() + "@" + objCounter
```
sun-rui: If we don't need the class name for instanceOf, we can remove the class name from the object ID to reduce socket traffic.

sun-rui:
@shivaram I added comments for minor issues:)

pkg/R/RDD.R
shivaram: So the reason we need a new env for a jobj is that we can only register finalizers for environments or external references pointers. Given that we need the finalizer for GC on the Java side, I am not sure we have a better solution for now

pkg/R/RDD.R
shivaram: The `ascii = TRUE` is not introduced in this PR and seems to go a long way back (introduced in https://github.com/amplab-extras/SparkR-pkg/commit/25a0bea75923abb1734afd3f947e6feb5df6d1d3)

I don't recall why we added it back then but I've opened a JIRA to investigate if things work fine without it
https://sparkr.atlassian.net/browse/SPARKR-179

pkg/R/RDD.R
shivaram: So the convention we have in the code is that we save a pointer to the JavaRDD in the R side. This is because we call methods like `collectPartitions` etc which are (a) sometimes only present in the JavaRDD and (b) the type signatures we depend on are the java versions. So we need the `asJavaRDD` to remain consistent with this

pkg/R/RDD.R
shivaram: Same as before. Lets investigate this is https://sparkr.atlassian.net/browse/SPARKR-179

pkg/R/RDD.R
shivaram: We can only call `partitionBy` on something that extends a JavaPairRDD (or has PairRDDFunctions implicit functions but AFAIK that only works in Scala).

pkg/R/utils.R
shivaram: Hmm thats a good point. Actually I am not sure we need a new RPC. Can't we do this dynamically ? For example

```
isInstance <- function(jobj) {
  cls <- SparkR:::callJMethod(jobj, "getClass")
  SparkR:::callJMethod(cls, "isInstance", jobj)
}
```

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
shivaram: More than the instanceOf check I added the class name just for debug / usability purposes. i.e if we have a jobj on the R side its good to know what it represents (something like a toString). We could override the print method to call getClass.getName or something like that. Any thoughts ?

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
shivaram: done

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SerializeJavaR.scala
shivaram: done

pkg/R/sparkRBackend.R
shivaram: This is a great idea - done

pkg/R/sparkR.R
shivaram: Yeah I added it as an argument to sparkR.init. Going forward we could also read an environment variable for this

pkg/R/RDD.R
shivaram: done

pkg/R/RDD.R
shivaram: done

pkg/R/sparkR.R
shivaram: Done

pkg/R/sparkR.R
shivaram: Done

pkg/R/sparkR.R
shivaram: Done

shivaram:
Thanks @sun-rui for the detailed review.  I've fixed issues in the new code but for existing code like `ascii = T` I think its better to look at them in a new JIRA.

There are a couple of comments like isInstanceOf that need to fixed -- And do let me know if I missed some thing.

pkg/R/utils.R
sun-rui: Suppose instanceOf is not called frequently, I think this way works

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
sun-rui: It's ok to override the print method to print the class name of a jobj.

sun-rui:
Thanks @shivaram. I replied to the comments on instanceOf.

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
shivaram: I added a `print.jobj` which prints the class name, id.

pkg/R/utils.R
shivaram: We only seem to use it in `collect` which is not that frequently used. I added a utility function to sparkRBackend.R

shivaram:
Thanks @sun-rui -- I fixed the isInstanceOf and jobj id. Let me know if it looks okay.

sun-rui:
@shivaram  LGTM:)

shivaram:
Thanks @sun-rui -- I'm going to leave this open for a couple of more days for @concretevitamin and for any other comments.

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackend.scala
```diff
@@ -0,0 +1,99 @@
...
+
+import io.netty.bootstrap.ServerBootstrap
+import io.netty.channel.ChannelFuture
+import io.netty.channel.ChannelInitializer
+import io.netty.channel.ChannelOption
+import io.netty.channel.EventLoopGroup
+import io.netty.channel.nio.NioEventLoopGroup
+import io.netty.channel.socket.SocketChannel
+import io.netty.channel.socket.nio.NioServerSocketChannel
+
```
concretevitamin: minor: remove empty line

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackend.scala
```diff
@@ -0,0 +1,99 @@
...
+import io.netty.handler.codec.bytes.ByteArrayDecoder
+import io.netty.handler.codec.bytes.ByteArrayEncoder
+import io.netty.handler.codec.LengthFieldBasedFrameDecoder
+
+/**
+ * Netty-based backend server that is used to communicate between R and Java.
+ */
+class SparkRBackend() {
```
concretevitamin: parenthesis here is not necessary

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SerializeJavaR.scala
```diff
@@ -0,0 +1,297 @@
+package edu.berkeley.cs.amplab.sparkr
+
+import scala.collection.mutable.HashMap
```
concretevitamin: Unnecessary import

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
```diff
@@ -0,0 +1,204 @@
...
+      if(!parameterWrapperType.isInstance(args(i))) {
+        return false
+      }
+    }
+    return true
+  }
+}
+
+/**
+ * Helper singleton that tracks Java objects returned to R.
```
concretevitamin: One general comment about nomenclature both in code and in comments: it seems to be the case that "Java objects" / "to/from Java" etc. is used a lot, which is a slight misnomer -- I think it'd be great if we can invest a bit in the nomenclature, and consistently refer to expressions like "JVM objects" / "between JVM and R runtime". This will help with the clarity of the code.

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
```diff
@@ -0,0 +1,204 @@
...
+
+  // TODO: This map should be thread-safe if we want to support multiple
+  // connections at the same time
+  val objMap = new HashMap[String, Object]
+
+  // TODO: We support only one connection now, so an integer is fine.
+  // Investiage using use atomic integer in the future.
+  var objCounter: Int = 0
+
+  def getObject(id: String): Object = {
```
concretevitamin: It's somewhat confusing to have this function co-exist with the (idiomatic) `get`. How about renaming this to `apply` to adhere to convention

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackend.scala
```diff
@@ -0,0 +1,99 @@
+package edu.berkeley.cs.amplab.sparkr
+
+import java.io.IOException
+import java.net.InetSocketAddress
+import java.util.concurrent.TimeUnit
+
+import io.netty.bootstrap.ServerBootstrap
+import io.netty.channel.ChannelFuture
+import io.netty.channel.ChannelInitializer
+import io.netty.channel.ChannelOption
```
concretevitamin: unused import

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackend.scala
```diff
@@ -0,0 +1,99 @@
...
+            // lengthFieldLength = 4
+            // lengthAdjustment = 0
+            // initialBytesToStrip = 4, i.e. strip out the length field itself
+            new LengthFieldBasedFrameDecoder(Integer.MAX_VALUE, 0, 4, 0, 4))
+          .addLast("decoder", new ByteArrayDecoder())
+          .addLast("handler", handler)
+      }
+    })
+
+    channelFuture = bootstrap.bind(new InetSocketAddress(port))
```
concretevitamin: Create the address at the beginning of the function? If any issues occur we'd catch it earlier.

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
```diff
@@ -0,0 +1,204 @@
...
+          case java.lang.Double.TYPE => classOf[java.lang.Double]
+          case java.lang.Boolean.TYPE => classOf[java.lang.Boolean]
+          case _ => parameterType
+        }
+      }
+      if(!parameterWrapperType.isInstance(args(i))) {
+        return false
+      }
+    }
+    return true
```
concretevitamin: More idiomatic to not use `return` here

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
```diff
@@ -0,0 +1,204 @@
...
+
+      if (parameterType.isPrimitive()) {
+        parameterWrapperType = parameterType match {
+          case java.lang.Integer.TYPE => classOf[java.lang.Integer]
+          case java.lang.Double.TYPE => classOf[java.lang.Double]
+          case java.lang.Boolean.TYPE => classOf[java.lang.Boolean]
+          case _ => parameterType
+        }
+      }
+      if(!parameterWrapperType.isInstance(args(i))) {
```
concretevitamin: space after if

pkg/R/RDD.R
```diff
@@ -1,6 +1,54 @@
...
+  obj
+}
+
+print.jobj <- function(jobj) {
+  cls <- callJMethod(jobj, "getClass")
+  name <- callJMethod(cls, "getName")
+  cat("Java ref type", name, "id", jobj$id, "\n", sep=" ")
+}
+
+cleanup.jobj <- function(e) {
```
concretevitamin: "e" might hint at "environment", perhaps better to use "obj" here

pkg/NAMESPACE
```diff
@@ -65,4 +65,5 @@ export(
        "setCheckpointDir"
       )
 export("sparkR.init")
+export("print.jobj")
```
concretevitamin: Is this exported for debugging use only?

pkg/R/RDD.R
```diff
@@ -1,6 +1,54 @@
...
+
+  if (.validJobjs[[objId]] == 0) {
+    rm(list=objId, envir=.validJobjs)
+    # NOTE: We cannot call removeJObject here as the finalizer may be run
+    # in the middle of another RPC. Thus we queue up this object Id to be removed
+    # and then run all the removeJObject when the next RPC is called.
+    .toRemoveJobjs[[objId]] <- 1
+  }
+}
+setOldClass("jobj")
```
concretevitamin: How about moving the jobj-related code out of RDD.R into either sparkRBackend.R or some other file?

pkg/R/sparkRBackend.R
```diff
@@ -0,0 +1,93 @@
...
+# callJMethod, callJStatic etc. instead of using this.
+#
+# If isStatic is true, objId contains className otherwise
+# it should contain a jobj returned previously by the backend
+invokeJava <- function(isStatic, objId, methodName, ...) {
+  if (!exists(".sparkRCon", .sparkREnv)) {
+    stop("No connection to backend found")
+  }
+
+  # If this is already isn't a removeJObject call
```
concretevitamin: remove "is already" -- got me a bit

pkg/R/sparkRBackend.R
```diff
@@ -0,0 +1,93 @@
...
+  if (!isRemoveMethod(isStatic, objId, methodName)) {
+    objsToRemove <- ls(.toRemoveJobjs)
+    if (length(objsToRemove) > 0) {
+      #cat("Clearing up objsToRemove\n")
```
concretevitamin: remove comment

pkg/R/sparkRBackend.R
```diff
@@ -0,0 +1,93 @@
...
+              removeJObject(e)
+            })
+      rm(list=objsToRemove, envir=.toRemoveJobjs)
+    }
+  }
+
+  rc <- rawConnection(raw(0), "r+")
+
+  writeBoolean(rc, isStatic)
+  # Write object id as string if it is static
```
concretevitamin: Not sure I get this comment (e.g. there's no check). Perhaps something like "If static call, objId encodes an object name as string; otherwise write out the object ID"

pkg/R/RDD.R
concretevitamin: nit: re-indent the `ascii = TRUE` line

pkg/R/serialize.R
```diff
@@ -0,0 +1,121 @@
...
+# character -> String
+# logical -> Boolean
+# double, numeric -> Double
+# raw -> Array[Byte]
+#
+# list[T] -> Array[T], where T is one of above mentioned types
+# environment -> Map[String, T], where T is a native type
+# jobj -> Object, where jobj is an object created in the backend
+
+writeObject <- function(con, object, withType = TRUE) {
```
concretevitamin: rename `withType` to `writeType`?

concretevitamin:
I'm still going over the patch (nice work!), but one general comment: do we have an idea of any performance change? The design doc lists no performance degradation as a goal.

pkg/R/RDD.R
```diff
@@ -1100,9 +1137,7 @@ setMethod("saveAsObjectFile",
             if (!rdd@env$serialized) {
               rdd <- reserialize(rdd)
             }
-            .jcall(getJRDD(rdd), "V", "saveAsObjectFile", path)
-            # NOTE: rJava doesn't check for exceptions if the return type is void
```
concretevitamin: Just wanna check with you that removing this line in the new patch does not get this regression.

pkg/R/RDD.R
```diff
@@ -1100,9 +1137,7 @@ setMethod("saveAsObjectFile",
...
-            .jcheck()
+            callJMethod(getJRDD(rdd), "saveAsObjectFile", path)
             # Return nothing
             invisible(NULL)
```
concretevitamin: While we're at it, we can make this `invisible` call take the last statement, as done in `setCheckpointDir`.

pkg/R/RDD.R
```diff
@@ -1130,9 +1165,7 @@ setMethod("saveAsTextFile",
               toString(x)
             }
             stringRdd <- lapply(rdd, func)
-            .jcall(getJRDD(stringRdd, dataSerialization = FALSE), "V", "saveAsTextFile", path)
-            # NOTE: rJava doesn't check for exceptions if the return type is void
-            .jcheck()
+            callJMethod(getJRDD(stringRdd, dataSerialization = FALSE), "saveAsTextFile", path)
             # Return nothing
             invisible(NULL)
```
concretevitamin: Similar `invisible` trick?

pkg/R/deserialize.R
```diff
@@ -1,9 +1,73 @@
...
+}
+
+readString <- function(con) {
+  stringLen <- readInt(con)
+  string <- readBin(con, raw(), stringLen, endian="big")
+  rawToChar(string)
+}
 
 readInt <- function(con) {
   readBin(con, integer(), n = 1, endian="big")
```
concretevitamin: Weird to have `arg = val` and `arg=val` at the same time...

pkg/R/deserialize.R
```diff
@@ -1,9 +1,73 @@
...
 }
 
+readDouble <- function(con) {
+  readBin(con, double(), n = 1, endian="big")
```
concretevitamin: Similar style issue as above

pkg/NAMESPACE
concretevitamin: I realize this is part of the S3 OO system that is used for printing a brief description of `jrdd` of an RDD. However do we need to export it?

concretevitamin:
Regarding my previous comment on perf: I casually ran `pi.R` for a few times and there seemed to be a significant speedup.

pkg/DESCRIPTION
```diff
@@ -5,10 +5,10 @@ Version: 0.1
 Date: 2013-09-09
 Author: Shivaram Venkataraman
 Maintainer: Shivaram Venkataraman <shivaram@cs.berkeley.edu>
+Imports: methods
```
concretevitamin: Is this required for things like `cleanup.jobj`?

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
```diff
@@ -0,0 +1,204 @@
...
+          case Some(o) => {
+            cls = Some(o.getClass)
+            obj = o
+          }
+        }
+      }
+
+      val args = readArgs(numArgs, dis)
+
+      val methods = cls.get.getMethods()
```
concretevitamin: remove paren

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
```diff
@@ -0,0 +1,204 @@
...
+      val selectedMethods = methods.filter(m => m.getName() == methodName)
```
concretevitamin: remove paren

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
```diff
@@ -0,0 +1,204 @@
...
+      if (selectedMethods.length > 0) {
+        val selectedMethod = selectedMethods.filter { x => 
+          matchMethod(numArgs, args, x.getParameterTypes())
```
concretevitamin: remove paren

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
```diff
@@ -0,0 +1,204 @@
...
+        }.head
+
+        val ret = selectedMethod.invoke(obj, args:_*)
+
+        // Write status bit
+        writeInt(dos, 0)
+        writeObject(dos, ret.asInstanceOf[AnyRef])
+      } else if (methodName == "<init>") {
+        // methodName should be "<init>" for constructor
+        val ctor = cls.get.getConstructors().filter { x =>
```
concretevitamin: remove paren

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
```diff
@@ -0,0 +1,204 @@
...
+          matchMethod(numArgs, args, x.getParameterTypes())
```
concretevitamin: remove paren

pkg/R/deserialize.R
```diff
@@ -1,9 +1,73 @@
-# Utility functions to serialize, deserialize etc.
+# Utility functions to deserialize objects from Java.
+
+# Type mapping from Java to R
+# 
+# void -> NULL
+# Int -> integer
+# String -> character
+# Boolean -> logical
+# Double -> numeric / double
```
concretevitamin: In fact we don't need a slash here -- double is used.

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SerializeJavaR.scala
```diff
@@ -0,0 +1,297 @@
...
+
+  // Methods to write out data from Java to R
+  //
+  // Type mapping from Java to R
+  //
+  // void -> NULL
+  // Int -> integer
+  // String -> character
+  // Boolean -> logical
+  // Double -> numeric / double
```
concretevitamin: double is used

pkg/R/sparkRBackend.R
```diff
@@ -0,0 +1,93 @@
...
+}
+
+# Call a static method on a specified className
+callJStatic <- function(className, methodName, ...) {
+  invokeJava(isStatic=TRUE, className, methodName, ...)
+}
+
+# Create a new object of the specified class name
+newJObject <- function(className, ...) {
+  invokeJava(isStatic=TRUE, className, methodName="<init>", ...)
```
concretevitamin: So I saw this point has been raised, but it seems possible to do this:
```
  def `<init>` = "hello"
```
Maybe use something like `__<init>` since "_" usually means reserved?

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
```diff
@@ -0,0 +1,204 @@
...
+    ctx.flush()
+  }
+
+  override def exceptionCaught(ctx: ChannelHandlerContext, cause: Throwable) {
+    // Close the connection when an exception is raised.
+    cause.printStackTrace()
+    ctx.close()
+  }
+
+  def handleMethodCall(isStatic: Boolean, objId: String, methodName: String,
```
concretevitamin: style: each argument on its own line, four space indent

pkg/R/RDD.R
concretevitamin: I had this question as well until I saw this explanation -- would be really good to add it as a comment!

pkg/R/serialize.R
```diff
@@ -0,0 +1,121 @@
...
+    numeric = writeDouble(con, object),
+    raw = writeRaw(con, object),
+    list = writeList(con, object),
+    jobj = writeString(con, object$id),
+    environment = writeEnv(con, object),
+    stop("Unsupported type for serialization"))
+}
+
+writeString <- function(con, value) {
+  writeInt(con, as.integer(nchar(value) + 1))
```
concretevitamin: Does `writeBin` automatically take care of writing the ending `\0`?

pkg/R/serialize.R
```diff
@@ -0,0 +1,121 @@
...
+}
+
+writeRaw <- function(con, batch) {
+  writeInt(con, length(batch))
+  writeBin(batch, con, endian="big")
+}
+
+writeType <- function(con, class) {
+  type <- switch(class,
+            integer = "i",
```
concretevitamin: indent is off

concretevitamin:
A weird behavior: 
```
> rdd <- parallelize(sc, list(c(1, 2), c(3, 4)), 2L)
> collect(rdd)
[[1]]
[1] 1 2

[[2]]
[1] 3 4

> rdd <- parallelize(sc, list(c(NULL, NULL), c(NULL, NULL)), 2L)
> collect(rdd)
[[1]]
NULL

[[2]]
NULL
```
I confirmed in Spark REPL that `sc.parallelize(Seq((None, None), (None, None))).collect` should work.

concretevitamin:
Regarding the above NULL issue: so this already doesn't work in upstream master. Any good idea on why this might be? If it's not deliberate / something obvious that I miss (quite possible at this hour) then we should file a JIRA to further investigate.

pkg/R/serialize.R
```diff
@@ -0,0 +1,121 @@
...
+
+# Used to pass arrays where all the elements are of the same type
+writeList <- function(con, arr) {
+  # All elements should be of same type
+  elemType <- unique(sapply(arr, function(elem) { class(elem) }))
+  stopifnot(length(elemType) <= 1)
+
+  # Write empty lists as strings ?
+  if (length(elemType) == 0) {
+    elemType <- "character"
```
concretevitamin: Using "character" which is a valid type name in R is somewhat confusing -- maybe use something like "dummyString"? The comment here is confusing too (the element type is arbitrarily chosen to be string).

pkg/R/serialize.R
```diff
@@ -0,0 +1,121 @@
...
+
+  writeInt(con, len)
+  if (len > 0) {
+    writeList(con, as.list(ls(env)))
+    vals <- lapply(ls(env), function(x) { env[[x]] })
+    writeList(con, as.list(vals))
+  }
+}
+
+writeArgs <- function(con, args) {
```
concretevitamin: Add a comment to say "differs from `writeList` in that `args` can have multiple types".

pkg/R/sparkR.R
```diff
@@ -6,14 +6,26 @@ sparkR.onLoad <- function(libname, pkgname) {
...
+    sc <- get(".sparkRjsc", envir=.sparkREnv)
+    callJMethod(sc, "stop")
+  }
+
+  if (exists(".sparkRCon", envir=.sparkREnv)) {
+    callJStatic("SparkRHandler", "stopBackend")
+    # Also close the connection and remove it from our env
+    conn <- get(".sparkRCon", .sparkREnv)
+    close(conn)
+    rm(".sparkRCon", envir=.sparkREnv)
```
concretevitamin: Since the whole R session is exiting, no need to rm this

pkg/R/sparkRClient.R
```diff
@@ -0,0 +1,39 @@
...
+  if (exists(".sparkRcon", envir=.sparkREnv)) {
+    cat("SparkRBackend client connection already exists\n")
+    return(get(".sparkRcon", envir=.sparkREnv))
+  }
+
+  con <- socketConnection(host=hostname, port=port, server=FALSE,
+                          blocking=TRUE, open="wb", timeout=timeout)
+
+  assign(".sparkRCon", con, envir=.sparkREnv)
+  get(".sparkRCon", envir=.sparkREnv)
```
concretevitamin: replace this line with `con`

pkg/R/deserialize.R
concretevitamin: Also mention Long is also converted to double? (and in another list)

pkg/R/sparkRBackend.R
```diff
@@ -0,0 +1,93 @@
+# Methods to call into SparkRBackend. 
+
+
+# Returns TRUE if object is an instance of given class
+isInstanceOf <- function(jobj, className) {
+  stopifnot(class(jobj) == "jobj")
```
concretevitamin: This function actually might lie on the critical path of some programs. Maybe worthwhile to remove this assert

pkg/R/sparkRBackend.R
```diff
@@ -0,0 +1,93 @@
...
+  cls <- SparkR:::callJStatic("java.lang.Class", "forName", className)
+  SparkR:::callJMethod(cls, "isInstance", jobj)
+}
+
+# Call a Java method named methodName on the object
+# specified by objId. objId should be a "jobj" returned
+# from the SparkRBackend.
+callJMethod <- function(objId, methodName, ...) {
+  stopifnot(class(objId) == "jobj")
+  invokeJava(isStatic=FALSE, objId, methodName, ...)
```
concretevitamin: Related to the critical path comment -- let's just pass in `objId$id` here

pkg/R/sparkRBackend.R
```diff
@@ -0,0 +1,93 @@
...
+      rm(list=objsToRemove, envir=.toRemoveJobjs)
+    }
+  }
+
+  rc <- rawConnection(raw(0), "r+")
+
+  writeBoolean(rc, isStatic)
+  # Write object id as string if it is static
+  # else check if its a jobj and write its id
+  if (isStatic) {
```
concretevitamin: Related to the critical path comment -- if we just pass `objId$id` in for the only case where `isStatic == false`, we can remove this branching.

pkg/R/sparkRBackend.R
```diff
@@ -0,0 +1,93 @@
...
+  stopifnot(class(jobj) == "jobj")
+  cls <- SparkR:::callJStatic("java.lang.Class", "forName", className)
+  SparkR:::callJMethod(cls, "isInstance", jobj)
+}
+
+# Call a Java method named methodName on the object
+# specified by objId. objId should be a "jobj" returned
+# from the SparkRBackend.
+callJMethod <- function(objId, methodName, ...) {
+  stopifnot(class(objId) == "jobj")
```
concretevitamin: Related to the critical path comment -- remove this assert?

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SerializeJavaR.scala
```diff
@@ -0,0 +1,297 @@
...
+          writeType(dos, "list")
+          writeBooleanArr(dos, value.asInstanceOf[Array[Boolean]])
+        }
+        case "[[B" => {
+          writeType(dos, "list")
+          writeBytesArr(dos, value.asInstanceOf[Array[Array[Byte]]])
+        }
+        case _ => {
+          // Handle array of objects
+          if (value.getClass().getName().startsWith("[L")) {
```
concretevitamin: remove two pairs of paren here

pkg/R/sparkRClient.R
```diff
@@ -0,0 +1,39 @@
...
+}
+
+# Launch the SparkR backend using a call to 'system'.
```
concretevitamin: s/'system'/`system2`

concretevitamin:
A general comment: since a new dependency is pulled in & things like sockets are being used, it'd be good idea to test this patch on all of Linux (which TravisCI uses), Mac (which we use), and Windows.

pkg/R/utils.R
```diff
@@ -16,51 +16,41 @@ convertJListToRList <- function(jList, flatten, logicalUpperBound = NULL, serial
...
+                 # one whole partition dense-packed together. We deserialize the
+                 # whole partition first, then cap the number of elements to be returned.
+
+                 # TODO: is it possible to distinguish element boundary so that we can
+                 # unserialize only what we need?
+                 if (!is.null(logicalUpperBound)) {
                    res <- head(res, n = logicalUpperBound)
+                 }
+               } else {
+                 # jElem is of a primitive Java type, is simplified to R's
```
concretevitamin: `jElem` now obsolete

concretevitamin: Also, is this case ever reached now?

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SerializeJavaR.scala
```diff
@@ -0,0 +1,297 @@
...
+  }
+
+  def readBoolean(in: DataInputStream) = {
+    val intVal = in.readInt()
+    if (intVal == 0) false else true
+  }
+
+  def readBytesArr(in: DataInputStream) = {
+    val len = readInt(in)
+    (0 until len).map(_ => readBytes(in)).toArray
```
concretevitamin: Just a note: this might be pretty slow compared to while loops. I guess for now it's okay to leave it as is.

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SerializeJavaR.scala
```diff
@@ -0,0 +1,297 @@
...
+    (0 until len).map(_ => readString(in)).toArray
+  }
+
+  def readList(dis: DataInputStream) = {
+    val arrType = readObjectType(dis)
+    arrType match {
+      case 'i' => readIntArr(dis)
+      case 'c' => readStringArr(dis)
+      case 'd' => readDoubleArr(dis)
+      case 'l' => readBooleanArr(dis)
```
concretevitamin: Why does this case call `readBooleanArr`?

pkg/DESCRIPTION
shivaram: No - I found this out while running a `R CMD check` and reading more about it, it looks like we should be using `Imports` to specify dependencies. http://r-pkgs.had.co.nz/description.html has more details (See under 'Other Dependencies').

To answer the other part of your question we need `methods` to use S4 classes.

pkg/NAMESPACE
shivaram: If we don't export it, it can't be used from the shell. The shell by default calls `print` when you just type in a variable. For example, without this export after creating a SparkContext, I get
```
> sc
<environment: 0x7fbc2306c0a8>
attr(,"class")
[1] "jobj"
```

After this export, I get
```
> sc
Java ref type org.apache.spark.api.java.JavaSparkContext id 0
```

pkg/R/RDD.R
shivaram: Yeah that makes sense. I think we have enough code that I am going to move it to `jobj.R`

pkg/R/RDD.R
shivaram: We catch all exceptions on the Java side when the method executes and return an error when we get an exception. So this regression shouldn't happen.

pkg/R/deserialize.R
shivaram: So I thought about this in more detail and it turns out we don't really have a defined style we use consistently. I checked the Google R style guide and a bunch of R doc pages (like `?read.csv`, `?paste` etc.) and it looks the `arg = value` style is more commonly used. I switched all usages in SparkR to have the space between `arg` and `val`. Let me know if you have any thoughts on this

pkg/R/serialize.R
shivaram: Yeah - so the string value in R contains a `\0` at the end (like strings in say C) and writeBin writes it out

pkg/R/deserialize.R
concretevitamin: +1 to consistently using `arg = val`

pkg/R/serialize.R
shivaram: It needs to be a valid typename as the Java side does a switch based on the typename. The trouble is that if somebody is passing an empty list we have no idea what the type should be on the java side. Some of this is only working right now as functions take in `Array[Object]` on the java side. 

This seems to be a general problem where if we don't have any elements we can't create the right array type on the Java side. I've written a better TODO for now

pkg/R/sparkRBackend.R
shivaram: Hmm that seems to be a scala feature that gets changed when compiling to bytecode. For example if we have a `def <init>="hello"` it gets compiled to `public java.lang.String $lessinit$greater();` -- And given that we can only call function names as they exist in bytecode, I don't think this should be a problem.

pkg/R/sparkRBackend.R
shivaram: That makes sense. I just changed `callJStatic` to use `obj$id` and removed the `if` statement here

pkg/R/serialize.R
concretevitamin: Sorry I wasn't clear before -- I was really suggesting something like this (which looks clearer to me):
```
...
    elemType <- class("dummy") # character type
...
writeType(con, elemType)
```
This doesn't exploit the (somewhat surprising) fact that a character literal can be directly used as a type in the switch statement of `writeType`.

pkg/R/sparkRBackend.R
concretevitamin: Sure, feel free to drop this

pkg/R/utils.R
shivaram: Yeah so this is the case that gets called when we have a `JavaRDD[String]` that gets `collect` called on it. In this case the java result is `ArrayList[String]` and when we do a `get` on the ArrayList, our deserializer automatically converts `String` to `character`

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SerializeJavaR.scala
shivaram: Ah nice catch - this is a bug. It should be 'b' as that is an array of `logical` in R

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
shivaram: So I thought about that - stylistically it didn't look correct to have an apply function on a static object (feels like a constructor). Just as an example
```
val obj = JavaObjectTracker("2")
```
looks like we are constructing a java object tracker rather than looking up things in it ?

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
shivaram: Okay - I renamed this to JVMObjectTracker and updated the comment above this

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
concretevitamin: That's a very valid point. I'm fine with either just overriding `get` or leave this as is.

shivaram:
@concretevitamin Thanks a lot for the detailed review. I've addressed most of the comments and will mark more things as done once I push out a change after testing.

- Regarding the performance, I am not very sure on how to test this. I did some minor profiling by printing time it takes per-RPC and overall it looks like it takes the same amount of time as before for big operations like `createSparkContext`, `getJRDD` etc. For smaller operations like `getClass` the RPC times are 1ms or lesser, so I don't think it should be a problem. And I think the `pi.R` speedup you are seeing has to do with us not loading `rJava` on the worker side as well (it used to get pulled in as a SparkR dependency). I will open a follow up JIRA to do some perf-benchmarks

- As for testing, I've run this on my machine (Mac), windows machine (EC2) and linux (Travis). I think @sun-rui has also tested this, but am not sure on what environments.

pkg/R/RDD.R
shivaram: Added now

pkg/R/RDD.R
shivaram: Done

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackend.scala
shivaram: Done

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackend.scala
shivaram: Done

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SerializeJavaR.scala
shivaram: Done

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackend.scala
shivaram: Done

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackend.scala
shivaram: Done

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
shivaram: Done

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
shivaram: Done

pkg/R/RDD.R
shivaram: Done

pkg/R/sparkRBackend.R
shivaram: Done

pkg/R/sparkRBackend.R
shivaram: Done

pkg/R/sparkRBackend.R
shivaram: This is removed now as the semantics of `invokeJava` changed. I also added a comment at the top of the function with something like "if static the id must be classname" etc.

pkg/R/serialize.R
shivaram: Done

pkg/R/RDD.R
shivaram: Done

pkg/R/RDD.R
shivaram: Done

pkg/R/deserialize.R
shivaram: Fixed now

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
shivaram: Done

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
shivaram: Done

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
shivaram: Done

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
shivaram: Done

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
shivaram: Done

pkg/R/deserialize.R
shivaram: Done both

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SerializeJavaR.scala
shivaram: Fixed now

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
shivaram: Done

pkg/R/serialize.R
shivaram: Ah got it - made it `class("somestring")` now

pkg/R/serialize.R
shivaram: Done

pkg/R/sparkRClient.R
shivaram: Done

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SerializeJavaR.scala
shivaram: Done

pkg/R/sparkRClient.R
shivaram: Done

pkg/R/sparkR.R
shivaram: Well technically this could be called whenever `sc` goes out of scope and not just on exit. So I'm going to leave this in just to be safe.

pkg/R/sparkRBackend.R
shivaram: Fixed now

pkg/R/serialize.R
shivaram: Fixed now

pkg/R/sparkRBackend.R
shivaram: Hmm - So I am not really sure having this assert makes a big performance difference. On the other hand it might be useful to debug cases where we pass in something that is not a jobj. I'd say lets keep it till we do some perf testing ?

shivaram:
One more note on the performance: I added some timing code around the call to `new RRDD` inside `getJRDD` as it is a function with a lot of arguments and one that is used pretty often. I then ran the unit tests and calculated the average time for this new call. The numbers on my laptop were that the old rJava code took 13.8ms while the newer code takes 5.6ms on average.

This was not a very rigorous test and I am not sure why things are faster, but I think should be safe to say this change shouldn't cause any performance degradation.

pkg/R/sparkRBackend.R
concretevitamin: sure

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SerializeJavaR.scala
```diff
@@ -0,0 +1,297 @@
...
+      case _ => throw new IllegalArgumentException(s"Invalid type $typeStr")
+    }
+  }
+
+  def writeObject(dos: DataOutputStream, value: Object) {
+    if (value == null) {
+      writeType(dos, "void")
+    } else {
+      value.getClass.getName match {
+        case "java.lang.String" => {
```
concretevitamin: These brackets after the guards are redundant (actually flagged in IntelliJ)

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SerializeJavaR.scala
```diff
@@ -0,0 +1,297 @@
...
+          writeType(dos, "list")
+          writeBooleanArr(dos, value.asInstanceOf[Array[Boolean]])
+        }
+        case "[[B" => {
+          writeType(dos, "list")
+          writeBytesArr(dos, value.asInstanceOf[Array[Array[Byte]]])
+        }
+        case _ => {
+          // Handle array of objects
+          if (value.getClass.getName.startsWith("[L")) {
```
concretevitamin: Can be simplified by
```
case otherName =>
  if (otherName.startsWith("[L")) ...
```

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SerializeJavaR.scala
```diff
@@ -0,0 +1,297 @@
...
+  def readMap(in: DataInputStream): java.util.Map[Object, Object] = {
+    val len = readInt(in)
+    if (len > 0) {
+      val keysType = readObjectType(in)
+      val keysLen = readInt(in)
+      val keys = (0 until keysLen).map(_ => readTypedObject(in, keysType))
+
+      val valuesType = readObjectType(in)
+      val valuesLen = readInt(in)
+      val values = (0 until len).map(_ => readTypedObject(in, valuesType))
```
concretevitamin: Should be `(0 until valuesLen)`

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
```diff
@@ -0,0 +1,208 @@
...
+
+    // First bit is isStatic
+    val isStatic = readBoolean(dis)
+    val objId = readString(dis)
+    val methodName = readString(dis)
+    val numArgs = readInt(dis)
+
+    if (objId == "SparkRHandler") {
+      methodName match {
+        case "stopBackend" => {
```
concretevitamin: redundant braces

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
```diff
@@ -0,0 +1,208 @@
...
+          writeInt(dos, 0)
+          writeType(dos, "void")
+          server.close()
+        }
+        case "rm" => {
```
concretevitamin: redundant braces

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
```diff
@@ -0,0 +1,208 @@
...
+          try {
+            val t = readObjectType(dis)
+            assert(t == 'c')
+            val objToRemove = readString(dis)
+            JVMObjectTracker.remove(objToRemove)
+            writeInt(dos, 0)
+            writeObject(dos, null)
+          } catch {
+            case e: Exception => {
```
concretevitamin: redundant braces

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
```diff
@@ -0,0 +1,208 @@
...
+      dos: DataOutputStream) {
+    var obj: Object = null
+    var cls: Option[Class[_]] = None
+    try {
+      if (isStatic) {
+        cls = Some(Class.forName(objId))
+      } else {
+        JVMObjectTracker.get(objId) match {
+          case None => throw new IllegalArgumentException("Object not found " + objId)
+          case Some(o) => {
```
concretevitamin: redundant braces

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
```diff
@@ -0,0 +1,208 @@
...
+            cls = Some(o.getClass)
+            obj = o
+          }
+        }
+      }
+
+      val args = readArgs(numArgs, dis)
+
+      val methods = cls.get.getMethods
+      val selectedMethods = methods.filter(m => m.getName() == methodName)
```
concretevitamin: redundant paren?

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
```diff
@@ -0,0 +1,208 @@
...
+
+        val obj = ctor.newInstance(args:_*)
+
+        writeInt(dos, 0)
+        writeObject(dos, obj.asInstanceOf[AnyRef])
+      } else {
+        throw new IllegalArgumentException("invalid method " + methodName + " for object " + objId)
+      }
+    } catch {
+      case e: Exception => {
```
concretevitamin: redundant braces

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
```diff
@@ -0,0 +1,208 @@
...
+      parameterTypes: Array[Class[_]]): Boolean = {
+    if (parameterTypes.length != numArgs) {
+      return false
+    }
+
+    for (i <- 0 to numArgs - 1) {
+      val parameterType = parameterTypes(i)
+      var parameterWrapperType = parameterType
+
+      if (parameterType.isPrimitive()) {
```
concretevitamin: redundant paren

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SerializeJavaR.scala
```diff
@@ -0,0 +1,297 @@
...
+
+import scala.collection.JavaConversions._
+
+import java.io.DataInputStream
+import java.io.DataOutputStream
+
+/**
+ * Utility functions to serialize, deserialize objects to / from R
+ */
+object SerializeJavaR {
```
concretevitamin: Perhaps the conventional "SerDe" is a better name? (At least s/Java/JVM)

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SerializeJavaR.scala
```diff
@@ -0,0 +1,297 @@
...
+  def writeObject(dos: DataOutputStream, value: Object) {
+    if (value == null) {
+      writeType(dos, "void")
+    } else {
+      value.getClass.getName match {
+        case "java.lang.String" => {
+          writeType(dos, "character")
+          writeString(dos, value.asInstanceOf[String])
+        }
+        case "long" | "java.lang.Long" => {
```
concretevitamin: For the class names: will we ever run into things like `scala.Int`?

concretevitamin:
Hey @shivaram: thanks for the feedback on my questions. I finished my pass and the code in general LGTM.

sun-rui:
Thanks, @concretevitamin , very careful review. 

As for your comment about list(c(NULL, NULL), c(NULL, NULL)), 
in R:
> x<-c(NULL,NULL)
> x
NULL
> length(x)
[1] 0

 c(NULL, NULL) is still NULL, not something like (None, None) in scala. So the behaviour is right.

concretevitamin:
Ah I missed that. Thanks @sun-rui!

concretevitamin:
Also, just an idea: I find it a bit messy how the current code treats the various supported data types on the JVM side. A more principled approach might to exploit Scala's strong typing and introduce a small type hierarchy to let the type system do most of the work (think: a sealed class with each supported data type as a subclass, holding information like the characters that define them, conversions to/from objects of that type, etc.) instead of hand-rolling a bunch of dispatch logic in various places. This way, the logic is tightened, and it will be a central place where users/developers refer to to look for the list of supported types (currently it's specified in comment).

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SerializeJavaR.scala
shivaram: Removed all the brackets now - I usually like them as they show where the current case begins and ends, but I saw that code in Spark doesn't use them. Better to have consistent style

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SerializeJavaR.scala
shivaram: Done

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SerializeJavaR.scala
shivaram: Oh - nice catch ! Fixed now

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
shivaram: Removed

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
shivaram: Removed

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
shivaram: Removed

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
shivaram: Removed

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
shivaram: Removed

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
shivaram: Removed

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
shivaram: Removed

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SerializeJavaR.scala
shivaram: We will only get names that are associated with a Class type on the JVM. Scala ints, doubles etc. are mapped to the corresponding Java classes. For example on the scala shell
```
scala> val a = 4
scala> a.getClass.getName
res0: String = int
```

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SerializeJavaR.scala
shivaram: Hmm I renamed it to SerDeJVMR - I dont really like the name as there is no nice break between the words JVM and R. The main reason I had chosen Java gives us is with CamelCasing we can figure out the word boundaries, but I guess this is fine for now

concretevitamin: Yeah this a bit weird, how about just "SerDe"

concretevitamin: Or you can even just `object serde { ... }` -- it's not a class so I think it's fine to lower-case the first letter

shivaram: Hmm okay - I'll make it SerDe for now (the R and JVM part are kind of obvious I guess being in SparkR).

shivaram:
@concretevitamin @sun-rui Thanks again for the review. I'll merge after tests pass.

BTW The idea for refactoring the SerDe bits definitely sounds interesting and we should do try it out in a follow up PR. I was also thinking we could allow users to provide custom serializers -- something along the lines of what is supported by https://github.com/WayneSan/Pyrolite/blob/master/java/src/net/razorvine/pickle/Pickler.java . That will help in extending our SerDe to handle SparkSQL types like Date / Timestamp etc. in the future.

concretevitamin:
Congrats to @shivaram & @sun-rui for shipping this great feature!

shivaram:
@concretevitamin One worrying thing that I noticed is that the Jenkins, Travis tests slowed down now. This especially seems to be happening for the parallelize_collect tests. Could you take a look at https://sparkr.atlassian.net/browse/SPARKR-183 ?

PR #133, opened 2015-01-27
[SPARKR-183] [WIP] Add option to not reuse connection to backend

shivaram:
This is one of the possible solutions to the slow RPC problem on Linux and so I've marked it as WIP. I'd also like to see it this improves Travis test times

PR #135, opened 2015-01-27
[WIP][SPARKR-183] Add support for setting TCP NODELAY using lsof

shivaram:
Another proposed solution which uses `lsof`

### DIFF
--- a/pkg/R/sparkRClient.R
+++ b/pkg/R/sparkRClient.R
@@ -11,9 +11,39 @@ connectBackend <- function(hostname, port, timeout = 60) {
   con <- socketConnection(host = hostname, port = port, server = FALSE,
                           blocking = TRUE, open = "wb", timeout = timeout)
 
+  # Set TCP NODELAY for sockets on Unix-like machines.
+  if (.Platform$OS.type == "unix") {
+    # Get the FD for this socket connection
+    sockfd <- getFD(hostname, port)
+    if (sockfd > 0) {
+      ret <- .Call("setTcpNoDelay", sockfd)
+      if (ret == 1) {
+        cat("Successfully set TCP NODELAY on ", sockfd, "\n")
+      }
+    }
+  }
+
   assign(".sparkRCon", con, envir = .sparkREnv)
   con
 }
+
+getFD <- function(hostname, port) {
+  lsofArg <- paste("-i@", hostname, ":", port, sep="")
+  tryCatch({
+     lsofOut <- system2("lsof", lsofArg, stdout = T)
+     # Get the row corresponding to R's pid
+     # and then get the 4th column which is the FD
+     fdStr <- strsplit(lsofOut[grep(Sys.getpid(), lsofOut)], "[ ]+")[[1]][4]
+     return(as.integer(gsub("[a-zA-Z]+", "", fdStr)))
+  }, warning = function(warn) {
+     cat("Could not find socket descriptor\n")
+     return(-1)
+  }, error = function(err) {
+     cat("Could not find socket descriptor\n")
+     return(-1)
+  })
+}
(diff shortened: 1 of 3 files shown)


pkg/R/sparkRClient.R
```diff
@@ -11,10 +11,40 @@ connectBackend <- function(hostname, port, timeout = 60) {
...
                           blocking = TRUE, open = "wb", timeout = timeout)
 
+  # Set TCP NODELAY for sockets on Unix-like machines.
+  if (.Platform$OS.type == "unix") {
+    # Get the FD for this socket connection
+    sockfd <- getFD(hostname, port)
+    if (sockfd > 0) {
+      ret <- .Call("setTcpNoDelay", sockfd)
+      if (ret == 1) {
+        cat("Successfully set TCP NODELAY on ", sockfd, "\n")
```
concretevitamin: "Successfully set TCP NODELAY on  4 " ([Travis log](https://travis-ci.org/amplab-extras/SparkR-pkg/builds/48546586)) -- `sockfd` getting printed as a number is for debugging only? When I saw this message I initially thought it's "on 4" or something...

pkg/R/sparkRClient.R
```diff
@@ -11,10 +11,40 @@ connectBackend <- function(hostname, port, timeout = 60) {
   con <- socketConnection(host = hostname, port = port, server = FALSE,
                           blocking = TRUE, open = "wb", timeout = timeout)
 
+  # Set TCP NODELAY for sockets on Unix-like machines.
```
concretevitamin: Briefly explain the purpose is to mitigate SPARKR-183?

concretevitamin:
Given that both this patch and #133 seem to mitigate the issue a bit, do you have a preference over which to use?

shivaram:
I think this patch is a bit better in the sense that it actually addresses the problem. The only down side is that we have some more C code to maintain and we have to test if the `lsof` parsing doesn't break over time etc.

Any other pros and cons ? Also @sun-rui, do you have any  thoughts on this vs. #133 ?

concretevitamin:
I'm +1 on this too over #133 with the same reason (the other patch exploits a black-box observation of TCP implementation-dependent behaviors).

pkg/R/sparkRClient.R
shivaram: Done

pkg/R/sparkRClient.R
shivaram: Yeah it was mostly for debugging. I just removed these 3 lines now.

sun-rui:
I thought about using unix domain socket between the R and backend to avoid TCP stack overhead and also this issue, but notice that both R and Netty now don't support unix domain socket.

I preferred using TCP_NODELAY to reusing connection. However, it is a little bit hard to set TCP_NODELAY, because R does not provide formal API to get the socket fd for a socket connection. Using "lsof" is tricky as I am not sure if "lsof" is guarrantted to exist in all linux distributions. Also Windows platform is not supported.

After investigation, I found that it is recommended to avoid write-write-read pattern in socket programming before resorting to TCP_NODELAY. So I did a change to invokeJava, which construct the whole request message to send it once. And the test performance data I collected as below (all performance data was tested without Zhongheng's fix for collect()):

Original: 13m2.717s
Shivaram's TCP_NODELAY fix: 2m50.895s
My fix: 2m55.187s

there is basically no performance difference between Shivaram's TCP_NODELAY fix and My fix. So I recommend to use my fix as it is more general.

I sent a PR @https://github.com/amplab-extras/SparkR-pkg/pull/140  for your review.

shivaram:
Closing this as #140 is merged

PR #137, opened 2015-01-28
Fix documentation for includePackage

shivaram:
The example code was missing `sc` in the call to `includePackage`

cc @karthik


concretevitamin:
LGTM

PR #139, opened 2015-01-29
[SPARKR-185] Fix SparkR backend to exit in more cases

shivaram:
This change has two fixes
1. When the workspace is saved (from R or RStudio) the backend connection seems to be closed before the finalizer is run. In such cases we reopen the connection and stop the backend
2. With RStudio when R is restarted, there are port-conflicts which appear due to a race condition between the JVM and rsession restart. This change adds a 1 sec sleep to avoid this race.

cc @piccolbo

### DIFF
--- a/pkg/R/sparkR.R
+++ b/pkg/R/sparkR.R
@@ -11,21 +11,42 @@ sparkR.onLoad <- function(libname, pkgname) {
   .sparkREnv$assemblyJarPath <- assemblyJarPath
 }
 
-#' Stop this Spark context
-#' Also terminates the backend this R session is connected to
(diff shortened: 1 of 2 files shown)


cafreeman:
This is a much-needed improvement!  Looking forward to this getting pulled in.

piccolbo:
Do you want me to test from your fork or after the merge (simpler for me)?

shivaram:
I was just waiting to see if anybody had any code review comments before merging. Since this seems to be more critical, I'm going to merge this now and fix anything that comes up in a follow up.

piccolbo:
works for me

PR #140, opened 2015-01-29
[SPARKR-183] Fix the issue that parallelize collect tests are slow.

sun-rui:
Construct the whole request message to send it once, avoiding write-write-read pattern in case of Nagle's algorithm.

### DIFF
--- a/pkg/R/sparkRBackend.R
+++ b/pkg/R/sparkRBackend.R
@@ -72,18 +72,21 @@ invokeJava <- function(isStatic, objId, methodName, ...) {
   writeInt(rc, length(args))
   writeArgs(rc, args)
 
+  # Construct the whole request message to send it once,
+  # avoiding write-write-read pattern in case of Nagle's algorithm
   bytesToSend <- rawConnectionValue(rc)
+  close(rc)
+  rc <- rawConnection(raw(0), "r+")
+  writeInt(rc, length(bytesToSend))
+  writeBin(bytesToSend, rc)
+  bytesToSend <- rawConnectionValue(rc)
+  close(rc)
 
   conn <- get(".sparkRCon", .sparkREnv)
-  writeInt(conn, length(bytesToSend))
   writeBin(bytesToSend, conn)
 
   # TODO: check the status code to output error information
   returnStatus <- readInt(conn)
   stopifnot(returnStatus == 0)
-  ret <- readObject(conn)
-
-  close(rc) # TODO: Can we close this before ?
-
-  ret
+  readObject(conn)
 }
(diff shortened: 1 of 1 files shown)


pkg/R/sparkRBackend.R
```diff
@@ -72,18 +72,21 @@ invokeJava <- function(isStatic, objId, methodName, ...) {
   writeInt(rc, length(args))
   writeArgs(rc, args)
 
+  # Construct the whole request message to send it once,
+  # avoiding write-write-read pattern in case of Nagle's algorithm
```
concretevitamin: Would be good to add a reference to [this post by Nagle himself](http://developers.slashdot.org/comments.pl?sid=174457&cid=14515105) as well as SPARKR-183 to explain what's going on.

pkg/R/sparkRBackend.R
```diff
@@ -72,18 +72,21 @@ invokeJava <- function(isStatic, objId, methodName, ...) {
...
   bytesToSend <- rawConnectionValue(rc)
+  close(rc)
+  rc <- rawConnection(raw(0), "r+")
+  writeInt(rc, length(bytesToSend))
```
concretevitamin: I found reusing the variables a bit contrived. It's okay to reuse `rc`, but let's name the two payloads (the original `bytesToSend` and the final message `bytesToSend`) to different things?

concretevitamin:
For reference, [this post by Nagle](http://developers.slashdot.org/comments.pl?sid=174457&cid=14515105) recommends avoiding write-write-read at application-level instead of disabling the algorithm. 

I didn't read closely, but [this post](http://www.stuartcheshire.org/papers/NagleDelayedAck/) suggests that Windows not suffering from this issue is because of pure luck. So it sounds like doing an OS type check is not a sufficient workaround?

shivaram:
@sun-rui - This is a great find and this fix seems like a nice way to avoid the TCP delays. I tested this in a linux VM and it did seem to fix the issue.  So I'd vote for going with this change over #133 or #135 for now and we can revisit this if we find any further issues.

@concretevitamin - We don't need OS type checks with this change ? Also that post seems pretty old and I think MTU sizes have changed over the years (especially for the loopback interface we use in SpakR).

concretevitamin:
@shivaram I was referring to OS type check used in the other patch, so I'm also +1 on this fix too

pkg/R/sparkRBackend.R
sun-rui: fixed by adding a reference to http://en.wikipedia.org/wiki/Nagle%27s_algorithm

pkg/R/sparkRBackend.R
sun-rui: fixed

sun-rui:
fixed comments.

concretevitamin:
@sun-rui Can you rebase?

sun-rui:
rebased.

concretevitamin:
Thanks for the patch! Merging in.

PR #141, opened 2015-01-30
[SPARKR-140] Fix serialization tracking in pipelined RDDs

shivaram:
When creating a pipeline RDD, we need to check if the JavaRDD belonging to the parent is serialized.
Added a test case based on the JIRA 

cc @concretevitamin

### DIFF
--- a/pkg/R/RDD.R
+++ b/pkg/R/RDD.R
@@ -50,6 +50,7 @@ setMethod("initialize", "PipelinedRDD", function(.Object, prev, func, jrdd_val) 
   .Object@env$serialized <- prev@env$serialized
 
   # NOTE: We use prev_serialized to track if prev_jrdd is serialized
+  # prev_serialized is used during the delayed computation of JRDD in getJRDD
   .Object@prev <- prev
 
   isPipelinable <- function(rdd) {
--- a/pkg/R/RDD.R
+++ b/pkg/R/RDD.R
@@ -46,7 +46,10 @@ setMethod("initialize", "PipelinedRDD", function(.Object, prev, func, jrdd_val) 
   .Object@env$isCached <- FALSE
   .Object@env$isCheckpointed <- FALSE
   .Object@env$jrdd_val <- jrdd_val
+   # This tracks if jrdd_val is serialized
   .Object@env$serialized <- prev@env$serialized
+
+  # NOTE: We use prev_serialized to track if prev_jrdd is serialized
   .Object@prev <- prev
 
   isPipelinable <- function(rdd) {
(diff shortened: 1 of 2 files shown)


concretevitamin:
I think this is not a correct fix -- adding one more pipelined level breaks this:
```
rdd <- textFile(sc, "./README.md")
lengths <- lapply(rdd, function(x) { length(x) })
take(lengths, 5) # works
lengths10 <- lapply(lengths, function(x) { x + 10 })
take(lengths10, 2) # with this patch, works

lengths20 <- lapply(lengths10, function(x) { x + 20 })
take(lengths20, 2) # breaks
```

shivaram:
Ah - nice catch. I tried a different fix this time by introducing a new variable `prev_serialized`.
The contract that I tried to introduce is that in `PipelinedRDD` is that `serialized` tracks if `jrdd` is serialized or not while `prev_serialized` tracks if `prev_jrdd` is serialized or not.

pkg/inst/tests/test_textFile.R
```diff
@@ -122,3 +122,24 @@ test_that("textFile() on multiple paths", {
...
+test_that("Pipelined operations on RDDs created using textFile", {
+  fileName <- tempfile(pattern="spark-test", fileext=".tmp")
+  writeLines(mockFile, fileName)
+
+  rdd <- textFile(sc, fileName)
+
+  lengths <- lapply(rdd, function(x) { length(x) })
+  expect_equal(collect(lengths), list(1, 1))
+
+  lengthsPipelined <- lapply(lengths, function(x) { x + 10})
```
concretevitamin: super nit: space before "}"

pkg/inst/tests/test_textFile.R
```diff
@@ -122,3 +122,24 @@ test_that("textFile() on multiple paths", {
...
+  expect_equal(collect(lengthsPipelined), list(11, 11))
+
+  lengths30 <- lapply(lengthsPipelined, function(x) { x + 20 })
+  expect_equal(collect(lengths30), list(31, 31))
+
+  lengths20 <- lapply(lengths, function(x) { x + 20 })
```
concretevitamin: This case seems redundant?

pkg/R/RDD.R
```diff
@@ -46,7 +46,10 @@ setMethod("initialize", "PipelinedRDD", function(.Object, prev, func, jrdd_val)
   .Object@env$isCached <- FALSE
   .Object@env$isCheckpointed <- FALSE
   .Object@env$jrdd_val <- jrdd_val
+   # This tracks if jrdd_val is serialized
```
concretevitamin: Maybe also explain the purpose (since the pipelining logic is scattered around different places, it might become harder and harder to grasp as time goes by): "...used exactly once during the delayed computation of JRDD in `getJRDD`".

concretevitamin:
The solution looks good!

pkg/inst/tests/test_textFile.R
shivaram: Done

pkg/inst/tests/test_textFile.R
shivaram: Well I just added it to make sure it was picking up `prev_serialized` from the right place. I guess it is a bit redundant, but doesn't hurt

pkg/R/RDD.R
shivaram: Added a note on where `prev_serialized` is used

concretevitamin:
Thanks!

PR #142, opened 2015-01-30
[SPARKR-169] Suppress warnings from normalizePath

shivaram:


concretevitamin:
LGTM & verified for non-existent local files. Thanks!

PR #143, opened 2015-01-30
[SPARKR-181] Buffer stderr from R and return it on Exception

shivaram:
This change buffers the last 100 lines from R process and passes these lines back to the driver if we have an exception. This will help users debug why their tasks failed on the cluster

### DIFF
--- a/pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/RRDD.scala
+++ b/pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/RRDD.scala
@@ -32,36 +32,42 @@ private abstract class BaseRRDD[T: ClassTag, U: ClassTag](
     val pb = rWorkerProcessBuilder()
     val proc = pb.start()
 
-    startStderrThread(proc)
+    val errThread = startStderrThread(proc)
 
     val tempFile = startStdinThread(proc, parentIterator, split.index)
 
     // Return an iterator that read lines from the process's stdout
     val inputStream = new BufferedReader(new InputStreamReader(proc.getInputStream))
-    val stdOutFileName = inputStream.readLine().trim()
 
-    val dataStream = openDataStream(stdOutFileName)
+    try {
+      val stdOutFileName = inputStream.readLine().trim()
 
-    return new Iterator[U] {
-      def next(): U = {
-        val obj = _nextObj
-        if (hasNext) {
-          _nextObj = read()
+      val dataStream = openDataStream(stdOutFileName)
+
+      return new Iterator[U] {
+        def next(): U = {
+          val obj = _nextObj
+          if (hasNext) {
+            _nextObj = read()
+          }
+          obj
         }
-        obj
-      }
 
-      var _nextObj = read()
+        var _nextObj = read()
 
-      def hasNext(): Boolean = {
-        val hasMore = (_nextObj != null)
-        if (!hasMore) {
-          // Delete the temporary file we created as we are done reading it
-          dataStream.close()
-          tempFile.delete()
+        def hasNext(): Boolean = {
+          val hasMore = (_nextObj != null)
+          if (!hasMore) {
+            // Delete the temporary file we created as we are done reading it
+            dataStream.close()
+            tempFile.delete()
+          }
+          hasMore
         }
-        hasMore
       }
+    } catch {
+      case e: Exception =>
+        throw new SparkException("R computation failed with\n " + errThread.getLines())
     }
   }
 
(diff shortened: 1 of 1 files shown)


pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/RRDD.scala
```diff
@@ -312,6 +314,27 @@ private class StringRRDD[T: ClassTag](
   lazy val asJavaRDD : JavaRDD[String] = JavaRDD.fromRDD(this)
 }
 
+private class BufferedStreamThread(in: InputStream, name: String) extends Thread(name) {
+  val ERR_BUFFER_SIZE = 100
+  val lines = new Array[String](ERR_BUFFER_SIZE)
+  var lineIdx = 0
+  override def run() {
+    for (line <- Source.fromInputStream(in).getLines) {
```
concretevitamin: It seems to me this can be simplified to something like:
```
val output = Source.fromInputStream(in).getLines()
output.foreach(println)
lines = output.takeRight(ERR_BUFFER_SIZE)
```

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/RRDD.scala
```diff
@@ -312,6 +314,27 @@ private class StringRRDD[T: ClassTag](
...
+      lines(lineIdx) = line
+      lineIdx = (lineIdx + 1) % ERR_BUFFER_SIZE
+      System.err.println(line)
+    }
+  }
+
+  def getLines(): String = {
+    (0 until ERR_BUFFER_SIZE).filter { x =>
```
concretevitamin: With `takeRight` as suggested above we can avoid this kind of circular tracing. Also, is the filter on null looking for non-filled slots (in which case `takeRight` makes sure we don't need to), or is it just the lines returned can be null themselves (in which case the filter is still useful)?

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/RRDD.scala
shivaram: The trouble is `output` here is an `Iterator[String]`. If we want to take two passes on the iterator (one for println and one for takeRight) we'll need to convert it to an array and I don't want to buffer the entire `stderr` to an array.

concretevitamin: Ah, that makes sense.

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/RRDD.scala
```diff
@@ -32,36 +32,42 @@ private abstract class BaseRRDD[T: ClassTag, U: ClassTag](
...
     val tempFile = startStdinThread(proc, parentIterator, split.index)
 
     // Return an iterator that read lines from the process's stdout
     val inputStream = new BufferedReader(new InputStreamReader(proc.getInputStream))
-    val stdOutFileName = inputStream.readLine().trim()
 
-    val dataStream = openDataStream(stdOutFileName)
+    try {
+      val stdOutFileName = inputStream.readLine().trim()
+
```
concretevitamin: minor: remove empty line

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/RRDD.scala
```diff
@@ -312,6 +314,27 @@ private class StringRRDD[T: ClassTag](
   lazy val asJavaRDD : JavaRDD[String] = JavaRDD.fromRDD(this)
 }
 
+private class BufferedStreamThread(in: InputStream, name: String) extends Thread(name) {
+  val ERR_BUFFER_SIZE = 100
```
concretevitamin: Maybe make this a constructor argument

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/RRDD.scala
concretevitamin: Ok this comment doesn't apply given the discussion above.

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/RRDD.scala
shivaram: done

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/RRDD.scala
shivaram: done

concretevitamin:
Thanks @shivaram! Merging in.

PR #144, opened 2015-01-30
[SPARKR-176] Fix SparkR Rd files

shivaram:
This fixes a number of issues in SparkR man pages. The main changes are
1. Don't export or generate docs for PipelineRDD
2. Fix variable names for Filter, count to match base methods
3. Document missing arguments for sparkR.init, print.jobj etc.

With this change R CMD check --as-cran does not give any warnings

cc @cafreeman


concretevitamin:
LGTM. Thanks!

PR #145, opened 2015-01-31
[SPARKR-175] Refactor join code

lythesia:
Refactor out and reuse the shared logic & code of join-family functions.

### DIFF
--- a/pkg/R/utils.R
+++ b/pkg/R/utils.R
@@ -218,3 +218,53 @@ sortKeyValueList <- function(kv_list) {
   keys <- sapply(kv_list, function(x) x[[1]])
   kv_list[order(keys)]
 }
+
+## Utility function to generate compact R lists from grouped rdd 
+#  Used in Join-family functions
+genCompactLists <- function(tagged_list, cnull) {
+  len <- length(tagged_list)
+  num <- length(cnull)
+  lists <- list(vector("list", len), vector("list", len))
+  index <- list(1, 1)
+
+  for (x in tagged_list) {
+    tag <- x[[1]]
+    idx <- index[[tag]]
+    lists[[tag]][[idx]] <- x[[2]]
+    index[[tag]] <- idx + 1
+  }
+
+  len <- lapply(index, function(x) x-1)
+  for (i in (1:num)) {
+    if (cnull[[i]] && len[[i]] == 0) {
+      lists[[i]] <- list(NULL)
+    } else {
+      length(lists[[i]]) <- len[[i]]
+    }
+  }
+
+  lists
+}
+
+## Utility function to merge compact R lists
+#  Used in Join-family functions
+mergeCompactLists <- function(left, right) {
+  result <- list()
+  length(result) <- length(left) * length(right)
+  index <- 1
+  for (i in left) {
+    for (j in right) {
+      result[[index]] <- list(i, j)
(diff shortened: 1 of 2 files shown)


pkg/R/utils.R
```diff
@@ -218,3 +218,53 @@ sortKeyValueList <- function(kv_list) {
   keys <- sapply(kv_list, function(x) x[[1]])
   kv_list[order(keys)]
 }
+
+## Utility function to generate compact R lists from grouped rdd
```
concretevitamin: minor: follow the existing style using one # and one following space; similar below

pkg/R/utils.R
```diff
@@ -218,3 +218,53 @@ sortKeyValueList <- function(kv_list) {
...
+  index <- list(1, 1)
+
+  for (x in tagged_list) {
+    tag <- x[[1]]
+    idx <- index[[tag]]
+    lists[[tag]][[idx]] <- x[[2]]
+    index[[tag]] <- idx + 1
+  }
+
+  len <- lapply(index, function(x) x-1)
```
concretevitamin: minor: spaces around binary operator

pkg/R/utils.R
```diff
@@ -218,3 +218,53 @@ sortKeyValueList <- function(kv_list) {
...
+  for (i in (1:num)) {
```
concretevitamin: Do we need `num` as a variable? It's just constant 2 right?

pkg/R/utils.R
```diff
@@ -218,3 +218,53 @@ sortKeyValueList <- function(kv_list) {
   keys <- sapply(kv_list, function(x) x[[1]])
   kv_list[order(keys)]
 }
+
+## Utility function to generate compact R lists from grouped rdd 
+#  Used in Join-family functions
+genCompactLists <- function(tagged_list, cnull) {
```
concretevitamin: To increase clarity, could you add some comment on what these two arguments represent and what the function returns? Similarly for `mergeCompactLists` and `joinTaggedList`.

pkg/R/utils.R
```diff
@@ -218,3 +218,53 @@ sortKeyValueList <- function(kv_list) {
...
+    }
+  }
+  result
+}
+
+## Utility function to wrapper above two operations
+#  Used in Join-family functions
+joinTaggedList <- function(tagged_list, cnull) {
+  lists <- genCompactLists(tagged_list, cnull)
+  result <- mergeCompactLists(lists[[1]], lists[[2]])
```
concretevitamin: You can combine the two lines into one: `mergeCompactLists(lists[[1]], lists[[2]])`

concretevitamin:
This looks neat! I left some minor comments.

pkg/R/utils.R
lythesia: Basically I made this function more general to generate multiple compact lists each with specified tag(`1L`, `2L` ...), it may be useful in the future.

concretevitamin: I see. Do you have a pending feature in mind that can use this? Otherwise I'd say we should keep it simple & clear for now.

lythesia: Okay, I'll fix it. :)

concretevitamin:
LGTM. Thanks!

PR #146, opened 2015-01-31
Exclude hadoop client from Spark dependency

shivaram:
This fixes builds using Spark 1.2 inheriting the wrong hadoop-client version (especially with SBT).

Before this fix if you ran `SPARK_VERSION=1.2.0 ./install-dev.sh` you would see `hadoop-client-2.2.0` being included even though our hadoop-client version is set to 1.0.4. This is because spark-core depends on Hadoop Client 2.2.0 http://mvnrepository.com/artifact/org.apache.spark/spark-core_2.10/1.2.0


concretevitamin:
Verified on local machine. Thanks!

PR #147, opened 2015-02-01
Bunch of fixes for longer running jobs

shivaram:
1. Increase the timeout for socket connection to wait for long jobs
~~2. Add some profiling information in worker.R~~
3. Put temp file writes before stdin writes in RRDD.scala

### DIFF
--- a/pkg/R/sparkRClient.R
+++ b/pkg/R/sparkRClient.R
@@ -2,7 +2,7 @@
 
 # Creates a SparkR client connection object
 # if one doesn't already exist
-connectBackend <- function(hostname, port, timeout = 60) {
+connectBackend <- function(hostname, port, timeout = 6000) {
   if (exists(".sparkRcon", envir = .sparkREnv)) {
     cat("SparkRBackend client connection already exists\n")
     return(get(".sparkRcon", envir = .sparkREnv))
--- a/pkg/inst/worker/worker.R
+++ b/pkg/inst/worker/worker.R
@@ -1,7 +1,5 @@
 # Worker class
 
-begin <- proc.time()[3]
-
 # NOTE: We use "stdin" to get the process stdin instead of the command line
 inputConStdin  <- file("stdin", open = "rb")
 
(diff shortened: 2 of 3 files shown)


shivaram:
So one thing I am not sure about is including the profiling output in `stderr`. This makes SparkR very noisy when run in local mode. Any suggestions for how to make this information available on need ?

sun-rui:
@shivaram, I have an idea that allowing worker profiling is configurable.  The basic idea is to pass the worker profiling flag to the backend (java -D) and when createing RRDDs, capturing the flag into the instances. Then when RRDDs are launching R processes, command line option "--profile-worker" is passed to woker.R (an ENV var can be used instead of command line option?). 

Not sure if this idea works. You can refer to the WIP@https://github.com/sun-rui/SparkR-pkg/commits/sparkr-ec2-fixes.  It is not complete and it lacks the logic that captures the flag when creating RRDDs.

shivaram:
Thanks @sun-rui - The other thing I was thinking about is that we could follow the PySpark model and return these values to the RRDD (example https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/api/python/PythonRDD.scala#L123)

I think I'll remove the profiling from this PR as the other two bugs are important. We can come back to fix this in https://sparkr.atlassian.net/browse/SPARKR-188

concretevitamin:
Could you explain a bit on the partial state / temp file vs. stdout order issue?

shivaram:
So the issue is the following -- Lets call the temp fie `sparkr.temp`. Before this fix what was happening was that 
- Java process writes out name of the file `sparkr.temp` to stdin.
- R process reads in filename `sparkr.temp`.
- Java process would open `sparkr.temp` and write out all the metadata and then the length of data (say 10000) and then start writing data (lets say it takes 1 min to write this out)
- R process opens the same file and reads in metadata and the length 10000. It tries to read 10000 bytes from the file, but the file doesn't have all the bytes yet.

Right now we move the filename writing to the end and R blocks on stdin waiting for the filename.

concretevitamin:
Sounds good - thanks for the writeup.

PR #150, opened 2015-02-02
Have data frames handled appropriately by parallize

dputler:
Shivaram,
The issues we discussed concerning a data frame returning returning the value of TRUE when it is an argument to is.list().


pkg/R/context.R
```diff
@@ -85,11 +85,9 @@ parallelize <- function(sc, coll, numSlices = 1) {
   # TODO: bound/safeguard numSlices
   # TODO: unit tests for if the split works for all primitives
   # TODO: support matrix, data frame, etc
-  if (!is.list(coll)) {
-    if (!is.vector(coll)) {
-      message(paste("context.R: parallelize() currently only supports lists and vectors.",
-                    "Calling as.list() to coerce coll into a list."))
-    }
+  if ((!is.list(coll) && !is.vector(coll)) || is.data.frame(col)) {
```
shivaram: This should be `is.data.frame(coll)` (i.e. the variable has two `l`s)

sun-rui:
as.list() on a data frame converts it by columns. Does it make more sense to convert a data frame by rows?

pkg/R/context.R
dputler: Sorry, let me fix and resubmit.

dputler:
You have a point about this. One thing that complicates the situation is that parallelize currently creates a RDD, and a row based approach is really more appropriate for a SchemaRDD. We actually have in the works a write.SchemaRDD function that writes an R data frame to a SchemaRDD (each row is a Row RDD). The function has a numSlices option which partitions the SchemaRDD by rows.

 

Currently, when parallelize is used with a data frame, the return value from is.list() is TRUE, so the function attempts to directly parallelize the data frame, which results in the rows being sliced, but not in the expected way, and the functions frequently issues a warning. The current intended behavior of the function is to turn everything that isn’t a vector or a list into a list, which for a data frame handles things by columns. Because of how is.list() handles data frames, the function doesn’t produce the anticipated behavior. My proposed “fix” does, but whether that is really the desired behavior is another matter.

 

I’m going to hold-off on cleaning up my pull request until there is an agreement on the desired behavior. A related point is that currently if an R matrix is given as an argument to parallelize, as.list converts it into a list that consists of a set of single data elements, and since it there is no meta data conveyed, there isn’t really a clean way to reconstruct the original matrix once it is in Spark.

shivaram:
Hmm this is an interesting question -- Here is my take on it
- In the long run we should of course make `parallelize` create a SchemaRDD / DataFrame as Dan suggested.
- In the meanwhile, we could try and special case the data.frame and matrix cases to work more reasonably. For example, we could split the data.frame row-wise such that every partition is itself a data.frame and one can run `lapplyPartition` on it. I think we can do this if we call `split` with `nrow(coll)` instead of `length(coll)` in the same function. And we could of course try something like this for matrices.

However, I am not sure the benefits from doing this will be more than the possible confusion it causes users. Because if we expect `lapply` to work on RDDs as it works locally, then I think `as.list` is the best we can do ?

piccolbo:
The short term version is how plyrmr does it as well. For data frames use
split.data.frame and for matrices, surprisingly, split.data.frame.
split.matrix does something else. But I concur with Shivaram that adding
this behavior to lapply, which means list-apply, is not a great idea.
lapply is used all the time to iterate over the columns, e.g.
lapply(mtcars, class). Unfortunately, there isn't really an apply family
function for data frames, that's why plyr and dplyr are so popular, besides
other reasons. The apply function can run on data.frames, but it will
return a matrix.

dputler:
If the goal is to parallelize things so that when lapply is called against them, the behavior matches traditional R, then everything that is not either a vector or a list should be coerced to being a list before they are parallelized to make their behavior consistent with traditional R.

 

Replicating plyrmr functionality seems very worthwhile, but seems like a slightly different issue.

 

My two cents.

piccolbo:
Replicating plyrmr seems a very bad idea, I hope I didn't even remotely
suggest it. It works on top of sparkR and can share contexts and data. I
don't see the point in copying it, just install and load.

shivaram:
Yeah I don't mind the coercing everything to a list as a short-term solution. We could also print a warning if we detect a data-frame saying that `NOTE: DataFrames will be get parallelized by columns` or something like that.

dputler:
I can do that.

 

One R factoid that can be used here is that is.list() returns TRUE for a list and a data frame, it turns out that is.vector() returns TRUE for a list, but FALSE for a data frame. So, the actual test just needs to be based on a call to is.vector().

piccolbo:
I wouldn't be so sure

> test = structure(list(data = 1), att = 1)
> is.vector(test)
[1] FALSE
> is.list(test)
[1] TRUE

dputler:
Thanks, it would appear that the addition of the attribute causes a reversal in the behavior of is.vector(), which I didn’t anticipate.

PR #154, opened 2015-02-04
[SPARKR-150] phase 1: implement sortBy() and sortByKey().

sun-rui:
This is phase 1 patch for SPARKR-150. It implements sortBy() and sortByKey().

Basically the rangePartitioner is the same as that in PySpark. It is simpler that the rangePartitioner in Spark, we can enhance it later if necessary.

Now add only two test cases. Will add more test cases during review.


pkg/R/RDD.R
```diff
@@ -1240,6 +1240,40 @@ setMethod("flatMapValues",
...
+#' collect(sortBy(rdd, function(x) { x })) # list (1, 2, 3)
+#'}
+setGeneric("sortBy", function(rdd, func, ascending, numPartitions) { standardGeneric("sortBy") })
+
+setClassUnion("missingOrLogical", c("missing", "logical"))
+#' @rdname sortBy
+#' @aliases sortBy,RDD,RDD-method
+setMethod("sortBy",
+          signature(rdd = "RDD", func = "function", 
+                    ascending = "missingOrLogical", numPartitions = "missingOrInteger"),
```
shivaram: Can't we just use default argument for `ascending` ? (Something like in `collect(..., flatten=TRUE)`). As far as I remember we needed the `missingOrInteger` because the default argument depended on the RDD for `numPartitions`.

pkg/R/RDD.R
```diff
@@ -1796,6 +1830,78 @@ setMethod("cogroup",
...
+#' sc <- sparkR.init()
+#' rdd <- parallelize(sc, list(list(3, 3), list(2, 2), list(1, 1)))
+#' collect(sortByKey(rdd)) # list (list(1, 1), list(2, 2), list(3, 3))
+#'}
+setGeneric("sortByKey", function(rdd, ascending, numPartitions) { standardGeneric("sortByKey") })
+
+#' @rdname sortByKey
+#' @aliases sortByKey,RDD,RDD-method
+setMethod("sortByKey",
+          signature(rdd = "RDD", ascending = "missingOrLogical", numPartitions = "missingOrInteger"),
```
shivaram: I know that `ascending` is in line with the RDD API, but I am wondering if `decreasing = FALSE` is better here as its more similar to R's `sort` function. Any thoughts ?

pkg/R/RDD.R
```diff
@@ -1796,6 +1830,78 @@ setMethod("cogroup",
...
+                                      function(i) {
+                                        j <- ceiling(length(samples) * i / numPartitions)
+                                        samples[j]
+                                      })
+              }
+            }
+
+            rangePartitionFunc <- function(key) {
+              partition <- 0
+
```
shivaram: You can add a `TODO` here that we should try to use binary search instead of linear search. We can try to implement a utils function to do that ?

shivaram:
Thanks @sun-rui for this change. This is looking pretty good ! I left some minor comments inline.

@concretevitamin -- it would be great if you could also take a look

pkg/R/RDD.R
sun-rui: Yes, I did try using default arguments in setMethod. But seems there is bug or something I don't know that make the default arguments not work in setMethod. This actually is a workaround which seems a little bit ugly. I will do more investigation to see if I can use default arguments.

pkg/R/RDD.R
sun-rui: I think consistency with RDD API is more important. It won't confusing programmers with RDD API in mind.

pkg/R/RDD.R
sun-rui: OK. I will add a TODO

pkg/R/RDD.R
shivaram: I think the trick is we can't dispatch with the optional argument. i.e. we can't have `ascending` in the `setGeneric` method. Also we might need to make it the last argument to get it to work

pkg/R/RDD.R
```diff
@@ -1796,6 +1830,79 @@ setMethod("cogroup",
...
+#' @param rdd An (k, v) pair RDD to be sorted.
+#' @param ascending A flag to indicate whether the sorting is ascending or descending.
+#' @param numPartitions Number of partitions to create.
+#' @return An RDD where all (k, v) pair elements are sorted.
+#' @rdname sortByKey
+#' @export
+#' @examples
+#'\dontrun{
+#' sc <- sparkR.init()
+#' rdd <- parallelize(sc, list(list(3, 3), list(2, 2), list(1, 1)))
```
concretevitamin: I think it'd be more educational to users if we, say, have key-val inversions (e.g. (3, 1), ...).

pkg/R/RDD.R
concretevitamin: +1 on keeping the RDD API

pkg/R/RDD.R
```diff
@@ -1796,6 +1830,79 @@ setMethod("cogroup",
...
+            
+            if (numPartitions > 1) {
+              rddSize <- count(rdd)
+              # constant from Spark's RangePartitioner
+              maxSampleSize <- numPartitions * 20
+              fraction <- min(maxSampleSize / max(rddSize, 1), 1.0)
+              
+              samples <- collect(keys(sampleRDD(rdd, FALSE, fraction, 1L)))
+              
+              # Note: the built-in R sort() function only atomic vectors
```
concretevitamin: missing "...works on..."?

pkg/R/RDD.R
```diff
@@ -1796,6 +1830,79 @@ setMethod("cogroup",
...
+                                        j <- ceiling(length(samples) * i / numPartitions)
+                                        samples[j]
+                                      })
+              }
+            }
+
+            rangePartitionFunc <- function(key) {
+              partition <- 0
+              
+              // TODO: Use binary search instead of linear search, similar with Spark
```
concretevitamin: Use # instead of // ?

pkg/inst/tests/test_rdd.R
```diff
@@ -254,6 +254,17 @@ test_that("keyBy on RDDs", {
   expect_equal(actual, lapply(nums, function(x) { list(func(x), x) }))
 })
 
+test_that("sortBy() on RDDs", {
```
concretevitamin: Would be good to test on a non-identity function.

pkg/inst/tests/test_rdd.R
```diff
@@ -373,3 +384,17 @@ test_that("fullOuterJoin() on pairwise RDDs", {
...
+  actual <- collect(sortedRdd)
+  numPairs <- lapply(nums, function(x) { list (x, x) })
+  expect_equal(actual, sortKeyValueList(numPairs, decreasing = TRUE))
+
+  rdd2 <- parallelize(sc, sort(nums, decreasing = TRUE), 2L)
+  numPairsRdd2 <- map(rdd2, function(x) { list (x, x) })
+  sortedRdd2 <- sortByKey(numPairsRdd2)
+  actual <- collect(sortedRdd2)
+  expect_equal(actual, numPairs)
+})
```
concretevitamin: newline at EOF

pkg/inst/tests/test_rdd.R
```diff
@@ -373,3 +384,17 @@ test_that("fullOuterJoin() on pairwise RDDs", {
   expect_equal(sortKeyValueList(actual),
                sortKeyValueList(list(list("a", list(1, NULL)), list("b", list(2, NULL)), list("d", list(NULL, 4)), list("c", list(NULL, 3)))))
 })
+
+test_that("sortByKey() on pairwise RDDs", {
+  numPairsRdd <- map(rdd, function(x) { list (x, x) })
```
concretevitamin: Would be good to have string keys?

pkg/R/RDD.R
concretevitamin: Does R have some binary search built in that we can use as an one-liner? If it does then it'd be good to add it.

concretevitamin:
I did a pass and the patch looks good - thanks @sun-rui. A comment on testing: it'd be good to test some boundary cases where there's only one element and/or one partition in an RDD, etc.

pkg/R/RDD.R
sun-rui: fixed

pkg/R/RDD.R
sun-rui: fixed

pkg/R/RDD.R
sun-rui: fixed

pkg/inst/tests/test_rdd.R
sun-rui: fixed. Use function (x) { x * x }.

pkg/inst/tests/test_rdd.R
sun-rui: fixed

pkg/inst/tests/test_rdd.R
sun-rui: fixed

sun-rui:
thanks @concretevitamin. Add more tests on boundary cases.

pkg/R/RDD.R
sun-rui: Use default arguments for sortBy() and sortKeyBy().

pkg/R/RDD.R
sun-rui: R does not have built-in binary search function. While there are two third-party packages, like gtools and data.table, providing binary search functions, they are GPL licensed.

concretevitamin:
@sun-rui could you rebase?

sun-rui:
Merged with upstream master.

shivaram:
Thanks @sun-rui - LGTM. @concretevitamin - Any further comments ?

concretevitamin:
LGTM - merging this.

PR #156, opened 2015-02-05
export sparkR.stop 

oscaroboto:


shivaram:
@oscaroboto -- `sparkR.stop` should be automatically called when R exits, so I am not sure we need to export it. Is there a specific bug you ran into ?

cafreeman:
I've run into issues where Rstudio can seemingly get "disconnected" from the backend so that I can't initialize a spark context even after quitting and restarting R several times. In those cases I have to manually kill the backend process from the command line and restart R.

That said, this is a rare case and is likely the result of me building the package over and over, but I still like the idea of being able to kill the backend from within R.  If you can manually initialize the spark context from inside your IDE, why not also expose the ability to manually kill it?

oscaroboto:
I also have run into the issue mentioned by @cafreeman regarding getting "disconnected". In which case I need to go to command line and kill the java pid manually. I think being able to kill the sc from within R will help. Furthermore, I would like to have the ability to kill and restart the sc from within an R session so that I don't have to close and reopen Rstudio while debugging and selecting cluster parameters such as number of instances/cores and spark worker memory.

README.md
```diff
@@ -84,7 +84,10 @@ pass the variable `spark.executor.memory` to the SparkContext constructor.
     sc <- sparkR.init(master="spark://<master>:7077",
                       sparkEnvir=list(spark.executor.memory="1g"))
 
+Finally, to stop the cluster run
 
+    sparkR.stop()
+    
 ## Examples, Unit tests
```
sun-rui: Suggest the following wording:
sparkR.stop() can be invoked to terminate a SparkContext created previously via sparkR.init(). Then you can call sparR.init() again to create a new SparkContext that may have different configurations.

pkg/R/sparkR.R
```diff
@@ -15,7 +15,7 @@ sparkR.onLoad <- function(libname, pkgname) {
 # backend and FALSE otherwise
 connExists <- function(env) {
   tryCatch({
-    exists(".sparkRCon", envir = env) && isOpen(env[[".sparkRCon"]])
+    !exists(".sparkRCon", envir = env) || (exists(".sparkRCon", envir = env) && isOpen(env[[".sparkRCon"]]))
```
sun-rui: Why this change? seems it is not necessary.

pkg/R/sparkR.R
```diff
@@ -23,7 +23,7 @@ connExists <- function(env) {
 
 # Stop the Spark context.
 # Also terminates the backend this R session is connected to
-sparkR.stop <- function(env) {
+sparkR.stop <- function(env = .sparkREnv) {
```
sun-rui: Don't add the default parameter. sparkR.stop is a finalizer that must be passed a env parameter by GC.

sun-rui:
I am not using RStudio, so I don't know why RStudio brutally restart a R session without cleaning up (i.e., sparkR.stop() finalizer is called).

However, I think it makes sense to expose sparkR.stop(). As pointed out by @oscaroboto, a user may want to stop the current SparkContext and create a new SparkContext with different configurations. Also it makes sense if in the future we may support multiple SparkContexts.

I left some minor comments on the change.

pkg/R/sparkR.R
oscaroboto: @sun-rui  This was necessary because if sparkR.stop() is inadvertently called multiple times or called before exiting R an error occurs, which I paste below. The error stems from the fact that sparkR.stop() removes ".sparkRCon"

>sc <- sparkR.init("local")
>sparkR.stop()
>sparkR.stop()
Error in socketConnection(host = hostname, port = port, server = FALSE,  : 
  cannot open the connection 
3 socketConnection(host = hostname, port = port, server = FALSE, 
    blocking = TRUE, open = "wb", timeout = timeout) at sparkRClient.R#11
2 connectBackend("localhost", .sparkREnv$sparkRBackendPort) at sparkR.R#34
1 sparkR.stop() 
In addition: Warning message:
In socketConnection(host = hostname, port = port, server = FALSE,  :
  localhost:12345 cannot be opened

pkg/R/sparkR.R
oscaroboto: The Default is needed here since it is not exposed

oscaroboto:
@sun-rui It will be really cool to support having multiple SparkContexts! Is there a Jira task for that?

sun-rui:
@oscaroboto, I depends on the support for multiple SparkContexts in Spark. As I know, it is not supported yet in Spark. @shivaram  If I were wrong, please correct me .

oscaroboto:
As you all know RStudio is not the most robust platform, which is why I think in some (maybe not so rare) cases it may shutdown and/or crash without cleaning up and hence not execute sparkR.stop() leading to a java pid running as a ghost process. I hate it when I get the bomb! (for those of you who know what I am talking about)

pkg/R/sparkR.R
sun-rui: My head was short-circuitted. I just deleted the comment but you still saw it:)

pkg/R/sparkR.R
```diff
@@ -36,14 +36,17 @@ sparkR.stop <- function(env) {
...
+    callJStatic("SparkRHandler", "stopBackend")
+    # Also close the connection and remove it from our env
+    conn <- get(".sparkRCon", env)
+    close(conn)
+    rm(".sparkRCon", envir = env)
+  }
+  
   # Finally, sleep for 1 sec to let backend finish exiting.
   # Without this we get port conflicts in RStudio when we try to 'Restart R'.
   Sys.sleep(1)
```
oscaroboto: I'm wondering if this Sys.sleep(1) should be placed within the if statement after line 47?

pkg/R/sparkR.R
oscaroboto: hahahaha

sun-rui:
yes, despite RStudio, it is still possible in practice that the R interperter losts connections with the backend for some reason, and the backend is not aware of this situation now. For robustness, the backend may check if the parent process exists at a predioc interval, if so it will terminate itself. However, I am not sure if it is worth to do so.

pkg/R/sparkR.R
sun-rui: yes, please move it into the if statement.

cafreeman: What would the benefit be?  If the goal is to have everything shut down completely before going to the next step, putting the sys.sleep outside of the function makes a lot of sense.  If nothing else, I think it that makes the code much more intuitive, and does so at little to no cost. 

What do you think?

pkg/R/sparkR.R
```diff
@@ -36,14 +36,17 @@ sparkR.stop <- function(env) {
   if (exists(".sparkRjsc", envir = env)) {
     sc <- get(".sparkRjsc", envir = env)
     callJMethod(sc, "stop")
+    rm(".sparkRjsc", envir = env)
   }
```
oscaroboto: I have another question. Do ".sparkRjsc" and ".sparkRCon" always exists simultaneously or is it possible for one to exist and not the other?
If so, these two if statements can be merged together into one if statement like, 
if (exists(".sparkRjsc", envir = env)&&exists(".sparkRCon", envir = env)) {...}

pkg/R/sparkR.R
oscaroboto: I was just thinking that if SparkR had been previously stopped and SparkR.stop() inadvertently called again or R exited the Sys.sleep(1) is not necessary and will slow the execution by 1 sec.

pkg/R/sparkR.R
oscaroboto: A rewrite of the two if statements could look like:
This also incorporates the Sys.sleep(1) within the if statement


  if (exists(".sparkRjsc", envir = env)&&exists(".sparkRCon", envir = env)) {
    cat("Stopping SparkR\n")
    sc <- get(".sparkRjsc", envir = env)
    callJMethod(sc, "stop")
    rm(".sparkRjsc", envir = env)

    callJStatic("SparkRHandler", "stopBackend")
    # Also close the connection and remove it from our env
    conn <- get(".sparkRCon", env)
    close(conn)
    rm(".sparkRCon", envir = env)
    # Finally, sleep for 1 sec to let backend finish exiting.
    # Without this we get port conflicts in RStudio when we try to 'Restart R'.
    Sys.sleep(1)
  }else{
    cat("SparkR Not Running: Call sparkR.init() to start SparkR")
  }

pkg/R/sparkR.R
cafreeman: Sounds good to me. Thanks :)

pkg/R/sparkR.R
sun-rui: yeah, sparkR.stop() may be called multiple times. But this change looks confusing.

I suggest to delete the connExists() function, and replace the line:
  if (!connExists(env)) {
with 
    if (exists(".sparkRCon", envir = env) && !isOpen(env[[".sparkRCon"]]) {

@shivaram , what do you think?

cafreeman: @sun-rui I think oscaroboto is on to the something, but the additional logic verges on tautology.  I think changing the connExists function is a more straightforward test for the same problem.

pkg/R/sparkR.R
shivaram: @oscaroboto - So technically we would like both to exist at the same time, but that doesn't always happen to be the case. Let me explain a bug that this code path was fixing (and this will also explain why connExists looks the way it does)

When `sparkR.stop` is called we want to do two things (a) stop the SparkContext and (b) stop the backend handler (Actually I am not that sure about (a), but lets postpone that for now). The problem is that to do either of those we need an active connection to the backend

- On exit from the R shell when one doesn't save the workspace, we usually have an active connection, so everything works fine i.e. `.sparkRCon` and `.sparkRjsc` both exit and the connection is valid. If you remove `.sparkRCon` from the env, this case will be idempotent i.e. you can call `sparkR.stop` many times without worry.
- On exit from the R shell (or RStudio) when one saves the workspace, R closes all active connections before calling the finalizer. So in this case we have a value in `.sparkRCon` but that connection is not valid. This is the main reason we have `connExists` right now is to catch an exception when we call `isOpen`.  Since we definitely want to stop the backend, we re-open a new connection if one does not exist. Now to make this idempotent what we need to do is to use `tryCatch` around the call to `connectBackend`. That way if the backend has already exited and we can't connect to it, we can just return from this function.
- Finally the `sleep(1)` should be only required in cases where we call `stopBackend` - so your call on that is correct. 

In summary I think the only thing that needs to change as a part of this PR is adding a `tryCatch` to `connectBackend` in this function and of course the `rm(.sparkRJsc)` above. Let me know if my analysis missed some use case.

oscaroboto: @shivaram Thank you for your reply, this does makes sense to me now. I added the  <tryCatch> to <connectBackend>, and added an error and warning message. 

I also changed the boolean statement in <connExists> to what it was originally.

Then I realized that <connectBackend> does a test if ".sparkRcon" exists, if it does it exists it returns ".sparkRcon" else it connects. I have modified <connectBackend> such that if ".sparkRcon" does exist but is not open then try to connect. If no connection exists throws a warning.

pkg/R/sparkR.R
```diff
@@ -23,30 +23,39 @@ connExists <- function(env) {
...
+sparkR.stop <- function(env = .sparkREnv) {
 
   if (!connExists(env)) {
     # When the workspace is saved in R, the connections are closed
     # *before* the finalizer is run. In these cases, we reconnect
     # to the backend, so we can shut it down.
-    connectBackend("localhost", .sparkREnv$sparkRBackendPort)
+    tryCatch({
+      connectBackend("localhost", .sparkREnv$sparkRBackendPort)
+    },error = function(err){
```
shivaram: Style nit: Space after `,` before `error` and `warning`. Also `Restart` can be just `restart` in the messages

pkg/R/sparkR.R
```diff
@@ -23,30 +23,39 @@ connExists <- function(env) {
...
+      cat("Error in Connection: Use sparkR.init() to Restart SparkR\n")
+    },warning = function(war){
+      cat("No Connection Found: Use sparkR.init() to Restart SparkR\n")
+    })
```
shivaram: we should probably add a `return(NULL)` here after printing the messages ? Or am I missing some exit path of the function ?

pkg/R/sparkR.R
```diff
@@ -23,30 +23,39 @@ connExists <- function(env) {
...
-  callJStatic("SparkRHandler", "stopBackend")
-  # Also close the connection and remove it from our env
-  conn <- get(".sparkRCon", env)
-  close(conn)
-  rm(".sparkRCon", envir = env)
-
-  # Finally, sleep for 1 sec to let backend finish exiting.
-  # Without this we get port conflicts in RStudio when we try to 'Restart R'.
-  Sys.sleep(1)
+  if(exists(".sparkRCon", envir = env)){
```
shivaram: I don't think we need this `if` statement anymore as `.sparkRCon` will always exist at this point. (`connectBackend` should have added it)

pkg/R/sparkRClient.R
```diff
@@ -4,8 +4,10 @@
 # if one doesn't already exist
 connectBackend <- function(hostname, port, timeout = 6000) {
   if (exists(".sparkRcon", envir = .sparkREnv)) {
-    cat("SparkRBackend client connection already exists\n")
-    return(get(".sparkRcon", envir = .sparkREnv))
+    if(isOpen(env[[".sparkRCon"]]))){
```
shivaram: This is a good catch ! Minor style nit: Space after `if` and before `{`

pkg/R/sparkR.R
oscaroboto: fixed

pkg/R/sparkR.R
oscaroboto: adding a return within tryCatch will only return tryCatch and continue with the rest of the function, even when error or warning are evoked. I have not though of a clever way around this. Hence the if(exists(".sparkRCon", envir = env)) is still necessary.

shivaram: Ah I see - Thanks. In that case can you move the `if (exists(".sparkRCon"))` above the calls to stopping the spark context ? If we don't have a connection we can't stop the spark context as well.

oscaroboto: I moved the if statement. @shivaram is this what you had in mind?

shivaram: Thanks @oscaroboto - that is exactly what I had in mind

shivaram:
Thanks @oscaroboto - The latest version mostly looks good and it seems to work as expected in R and RStudio on my laptop.

oscaroboto:
@shivaram Good to hear that!

shivaram:
Thanks @oscaroboto - I've squashed the commits and merged it into `master`. 

@cafreeman - Let us know if this helps with the problems you were running into.

oscaroboto:
@cafreeman So far so good! problem solved.

PR #158, opened 2015-02-06
[SPARKR-142][SPARKR-196] Adds cleanClosure() to capture the function dependencies. (Step 1 of solving UDF closure serialization issues).

hlin09:
This is an initial experimental try to solve the function/closure serialization the issue including Issue 142 and 196. It is just a minimal working version that passes the current tests. Please feel free to discuss whether it might be a reasonable direction and if so I will further clean up my PR.

The main idea behind this PR is to collect closure of each function in a `PipelinedRDD` and store the trace (lineage) of functions in a Accumulator inside the `PipelinedRDD`. The Accumulator will be serialized to worker side and the `computeFunc` is reconstructed into nested form for execution. 

The benefits of doing this is that the values of free variables that are not defined in the functions will be captured in time. Even if a user changes the values of these free variables before `PipelinedRDD` is materialized, they will be correctly stored in the Accumulator and serialized to the worker. 

One of the potential drawbacks of this PR is that the closures captured for each functions in the lineage might share some variables of the same value. I personally take it as a minor issue since 1) user rarely define free variables outside function scope, 2) normally these data size of these variables are small, in case of large variables, user should use other structures like broadcast vars, 3) it is solvable by removing the possible duplicated vars (maybe not worth doing it for now).

The PR also uses another way to get function dependencies. The previous `getDependencies()` get more than needed, and the dependency size grows exponentially! The approach in this PR (`clean.closure`) removes the unnecessaries the and keep the closure size in constant sizes as the program execution goes on.


pkg/R/RDD.R
```diff
@@ -526,6 +536,14 @@ setMethod("countByKey",
 setMethod("lapply",
           signature(X = "RDD", FUN = "function"),
           function(X, FUN) {
+            # Creats a new the closure (environment) for the FUN to capture
```
shivaram: Actually we can unify the code a little bit by implementing `lapply` using `lapplyPartitionsWithIndex`. In fact I think replacing the `PipelinedRDD` constructor here with `lapplyPartitionsWithIndex(X, func)` will do the trick.

pkg/R/utils.R
```diff
@@ -259,3 +267,40 @@ joinTaggedList <- function(tagged_list, cnull) {
...
+  if(nlen > 1) {
+    for(i in 1:nlen) {
+      closure.process(node[[i]], func, env)
+    }
+  } else if(nlen == 1) {
+    if(mode(node) == 'name') {
+      cnode <- as.character(node)
+      if(!cnode %in% names(as.list(args(func)))) { # Not a function argument
+        func.env <- environment(func)  
+        if(exists(cnode, envir=func.env, inherits=F)) {
```
shivaram: Hmm I am not able to figure out why `inherits` is false here ? All the default for `get` is `inherits=T` so it'll be good to set that to `F` if we don't want to look at parent environments. 

A comment above this code block will be very useful ~

pkg/R/RDD.R
```diff
@@ -70,10 +75,12 @@ setMethod("initialize", "PipelinedRDD", function(.Object, prev, func, jrdd_val)
       func(split, prev@func(split, iterator))
     }
     .Object@func <- pipelinedFunc
+    .Object@funcAccum <- cloneAccumulator(prev@funcAccum)
```
shivaram: So if we are going to go down this path of having a list of functions rather then we can get rid of the `pipelinedFunc` block above this.

However this brought up one more idea -- Do we really need a list of functions or can we somehow assign the environment of `pipelinedFunc` with something like `environment(pipelinedFunc) <- environment(prev@func)` ?
Basically what I am thinking about is that if we can do the pipelining on the worker side, is it possible to an equivalent pipeline on the driver side here.

pkg/R/RDD.R
```diff
@@ -526,6 +536,14 @@ setMethod("countByKey",
 setMethod("lapply",
           signature(X = "RDD", FUN = "function"),
           function(X, FUN) {
+            # Creats a new the closure (environment) for the FUN to capture
+            # free variables.
+            env.fun <- new.env(parent=.GlobalEnv)  
+            # need to add free variables to env.fun environment
+            clean.closure(FUN, env.fun)
```
shivaram: BTW this `clean.closure` is a great change !

shivaram:
@hlin09 -- Many apologies for the delay in reviewing this. Some high level comments in addition to my inline comments.

- It looks to me that this change actually has two mostly independent parts to it -- the first one being the `clean.closure` and the second being the accumulator storing functions. If that is the case we can probably split this into two PRs as the `clean.closure` might be easier to review and merge first

- It'll also be good to think of better test cases for both of these. For the `clean.closure` case can we have a wrapper function that lists the variables being captured ? If we had that then we could come up with some examples and add an assertion to make sure only those variables are captured etc.

pkg/R/RDD.R
hlin09: I agree. I will do it in a separate PR.

pkg/R/utils.R
hlin09: I was trying not to take the variables in the package environment (like base). Please take a look at my new commit.

hlin09:
@shivaram Thanks so much for comments. Please do not apologize. I will use the PR to create a cleanClosure and a separate one to handle functions. I will also update the PR description.

hlin09:
Closed and Transferred to PR #186 .

PR #159, opened 2015-02-10
[SPARKR-150] phase 2: implement takeOrdered() and top().

sun-rui:


pkg/R/RDD.R
```diff
@@ -1294,6 +1294,79 @@ setMethod("sortBy",
...
+      # R limitation: order works only on primitive types!
+      ord <- order(unlist(part, recursive = FALSE), decreasing = !ascending)
+      part[ord[1:num]]
+    } else {
+      part
+    }
+  }
+  
+  newRdd <- mapPartitions(rdd, partitionFunc)
+  take(sortBy(newRdd, function(x) { x }, ascending = ascending), num)
```
shivaram: Couldn't we just do a collect on `newRdd` and then take the top `num` ? It looks like the scala implementation is doing something similar [1] and it will save the cost of shuffling in `sortBy`

[1] https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/rdd/RDD.scala#L1237

sun-rui: @shivaram , yes, I did look at the scala implementation. But I thought that code may have scalability issue. If you take a small number of elements from a large data set, that's OK. If you take a large number of elements from a large data set, the driver may suffer OOM (number of partitions * Num elements need to be transferred to the driver node) or it takes time to sort on the single driver node. Just a thought, no performance data proving:)
@concretevitamin , do you have any comments?
I am fine to follow the scala implementation if you prefer that.

concretevitamin: @sun-rui In my experience I think `take` is most widely used as a data exploration tool, where users typically take a small number of elements for sanity check. (If a user wants a large amount of elements, JavaRDDLike has `collectPartitions()`.) So I'm +1 on following the Scala impl :)

nit: the anonymous function can be replaced by `identity`, should you need it later.

shivaram: One thing we could do to mimic the scala implementation is collect it partition at a time -- i.e. get the first partition, then get the second one and only keep the top `num` etc. This happens in scala automatically due to the bounded sized queue

sun-rui: Yeah, I did not realize that scala uses Bounded priority queue. OK I agree I will follow it. Actually pyspark also does so.

sun-rui:
fix comments.

pkg/R/RDD.R
shivaram: Unfortunately this doesn't quite fix it completely -- The `reduce` method in SparkR is implemented using a `collect` so this will get all `numPartitions * num` entries to the driver. FWIW I am fine with this approach for now.

It might be good to open a JIRA to fix `reduce` by calling `collect` on one partition at a time -- However that will reduce the amount of parallelism that we get and I'm not sure that is a good idea either (say a 1000 partition int RDD is being reduced using `+`, right now 1000 parts will be computed in parallel -- if we call collectPartition on one at a time then only 1 will be computed in parallel ?).

shivaram:
Alright, I think this looks good to me. We can track the `reduce` vs. `collect` issue separately -- @concretevitamin Any other comments ?

concretevitamin:
Instead of changing reduce()'s implementation to that way, we could do the
special handling only for these take() family functions? For now this patch
LGTM.

shivaram:
Yeah I think your idea is pretty good. We should just change the take related functions to work in the partition at a time manner. Created https://sparkr.atlassian.net/browse/SPARKR-199 to track this

PR #161, opened 2015-02-12
[SPARKR-150] Investigate if we can avoid using ascii=T in serialize calls

hqzizania:
The usage of "ascii=True" in 5 places has been removed. Tests and examples passed.


shivaram:
Thanks @hqzizania -- One thing I want to do is to test this with Windows, Mac, Linux to see if there are any platform specific gotchas. I'll try to test it today and then merge it

shivaram:
I just tested this on Windows as well and it works fine. (All unit tests pass etc.). Merging this.

hqzizania:
Sorry for my so late reply. I just tested this on Mac and Linux (Ubuntu) then.

PR #162, opened 2015-02-12
[SPARKR-159] Adds support for pipeRDD().

hlin09:
It seems that R does not support bi-direction pipe well. Here is an work around by using temp file to communicate between R and subprocess.


shivaram:
Thanks @hlin09 - This is very cool ! I'll take a closer look at this later today. Can you add [SPARKR-159] to the title of the PR ?

pkg/inst/tests/test_rdd.R
```diff
@@ -314,6 +314,18 @@ test_that("values() on RDDs", {
   expect_equal(actual, lapply(intPairs, function(x) { x[[2]] }))
 })
 
+test_that("pipeRDD() on RDDs", {
+  actual <- collect(pipeRDD(rdd, "cat"))
```
concretevitamin: Should we check for Unix OS type before running these tests?

hlin09: Thanks @concretevitamin for the catch! Should command "more" work better than "cat" in this case? Or I should check for OS type.

shivaram: I'd say adding a OS type check is fine. I don't think `more` will work on Windows anyways.

pkg/R/RDD.R
```diff
@@ -1374,6 +1374,48 @@ setMethod("top",
...
+
+#' @rdname pipeRDD
+#' @aliases pipeRDD,RDD,character-method
+setMethod("pipeRDD",
+          signature(rdd = "RDD", command = "character"),
+          function(rdd, command) {
+            func <- function(part) {
+              filename <- tempfile(pattern = "sparkr-pipe")
+              on.exit(unlink(filename))
+              pipeIn <- pipe(paste(command, ">", filename, sep=" "), "w")
```
shivaram: If we are going to go with output redirection, why not just use `system2` here ? We can set stdout to a variable in `system2` and stdin to the tempfile as you have now.

pkg/inst/tests/test_rdd.R
concretevitamin: +1

pkg/R/RDD.R
hlin09: Thanks @shivaram for mentioning `system2`. I guess I can remove tempfile by using `system2`.

shivaram:
@hlin09 - There seem to be a lot of files changed in this PR (probably due to some merge issue). Could you revert the files other than those in `RDD.R`, `NAMESPACE`, `test_rdd.R` and `pipeRDD.Rd` ?

pkg/inst/tests/test_rdd.R
sun-rui: "more" works on Windows. I think "more" is better than "cat" beause OS type check can be avoided.

pkg/R/RDD.R
```diff
@@ -1374,6 +1374,42 @@ setMethod("top",
             takeOrderedElem(rdd, num, FALSE)
           })
 
+#' Pipes elements to a forked externel process.
+#'
+#' The same as `pipe()' in Spark.
```
sun-rui: `pipe()' -> 'pipe()'

pkg/R/RDD.R
```diff
@@ -1374,6 +1374,42 @@ setMethod("top",
...
+#'
+#' @param rdd the RDD whose elements are piped to the forked externel process.
```
sun-rui: The first character of the description should be upper-case. 
the -> The

pkg/R/RDD.R
```diff
@@ -1374,6 +1374,42 @@ setMethod("top",
...
+#' @param command the command to fork an externel process.
```
sun-rui: the -> The

pkg/R/RDD.R
```diff
@@ -1374,6 +1374,42 @@ setMethod("top",
...
+#' @param ... other arguments passed to pipeRDD.
+#' @param env character vector of name=value strings to set environment variables.
+#' @return a new RDD created by piping all elements to a forked externel process.
```
sun-rui: a -> A

pkg/R/RDD.R
```diff
@@ -1374,6 +1374,42 @@ setMethod("top",
...
           })
 
+#' Pipes elements to a forked externel process.
+#'
+#' The same as `pipe()' in Spark.
+#'
+#' @param rdd the RDD whose elements are piped to the forked externel process.
+#' @param command the command to fork an externel process.
+#' @param ... other arguments passed to pipeRDD.
+#' @param env character vector of name=value strings to set environment variables.
```
sun-rui: character -> A character

pkg/R/RDD.R
```diff
@@ -1374,6 +1374,42 @@ setMethod("top",
...
+#' collect(pipeRDD(rdd, "cat")
+#' Output: c("1", "2", ..., "10")
+#'}
+setGeneric("pipeRDD", function(rdd, command, ...) { standardGeneric("pipeRDD") })
+
+#' @rdname pipeRDD
+#' @aliases pipeRDD,RDD,character-method
+setMethod("pipeRDD",
+          signature(rdd = "RDD", command = "character"),
+          function(rdd, command, env = character()) {
```
sun-rui: recommend the argument env to be a named list. This is to be consistent with those functions requires enviroment variables, for example , sparkR.init(). The named list can be converted to "name=value" strings as parameters to system2().

pkg/R/RDD.R
```diff
@@ -1374,6 +1374,42 @@ setMethod("top",
...
+            func <- function(part) {
+              trim.trailing.func <- function(x) {
```
sun-rui: Any reason for the trailing function?

pkg/R/RDD.R
```diff
@@ -1374,6 +1374,42 @@ setMethod("top",
...
+                sub("\\s+$", "", as.character(x))
```
sun-rui: Recommend to use toString() instead of as.character(). I think this is more representational and consistent with pyspark.
For example, if an RDD element x is list(1,2,3),
> x<-list(1,2,3)
> as.character(x)
[1] "1" "2" "3"
> toString(x)
[1] "1, 2, 3"

In python,
In [1]: x=[1,2,3]

In [2]: str(x)
Out[2]: '[1, 2, 3]'

pkg/R/RDD.R
```diff
@@ -1374,6 +1374,42 @@ setMethod("top",
...
+              }
+              input <- unlist(lapply(part, trim.trailing.func))
+              res <- system2(command, stdout = TRUE, input = input, env = env)
+              lapply(res, trim.trailing.func)
```
sun-rui: Is trimming necessary?

pkg/inst/tests/test_rdd.R
```diff
@@ -314,6 +314,20 @@ test_that("values() on RDDs", {
   expect_equal(actual, lapply(intPairs, function(x) { x[[2]] }))
 })
 
+test_that("pipeRDD() on RDDs", {
+  if (.Platform$OS.type == "unix") {
+    actual <- collect(pipeRDD(rdd, "cat"))
+    expected <- as.list(as.character(1:10))
+    expect_equal(actual, expected)
+  }
```
sun-rui: personally think, "more" is generic

pkg/R/RDD.R
```diff
@@ -1374,6 +1374,42 @@ setMethod("top",
...
+#' @rdname pipeRDD
+#' @export
+#' @examples
+#'\dontrun{
+#' sc <- sparkR.init()
+#' rdd <- parallelize(sc, 1:10)
+#' collect(pipeRDD(rdd, "cat")
+#' Output: c("1", "2", ..., "10")
+#'}
+setGeneric("pipeRDD", function(rdd, command, ...) { standardGeneric("pipeRDD") })
```
sun-rui: recommend to use:
setGeneric("pipeRDD", function(rdd, command, env = list()) { standardGeneric("pipeRDD") })

Use ... will allows to pass more than 3 arguments than needed.
You can refer to sortBy().

sun-rui:
@hlin09 , great contribution.  left some minor comments for discussion

pkg/inst/tests/test_rdd.R
hlin09: Either is fine with me. Let's see what @shivaram and @concretevitamin think about this.

pkg/R/RDD.R
hlin09: The function is used to trim the input fed to forked process so that system2 will not mess up the records by treating "\n" as a new line. Actually, I think only "\n" and "\r\n" needs to be removed.

pkg/R/RDD.R
hlin09: I am not sure. Probably not necessary, I was trying to follow the pyspark implementation.

pkg/inst/tests/test_rdd.R
shivaram: either is fine by me - if `more`works on Windows as well, then lets go with `more`

hlin09:
@shivaram Sorry for messing up with merges in my local repo. I am opening a new PR for this. Thanks @sun-rui for the detailed comments! I tried to fix them in the new PR, PTAL.

PR #165, opened 2015-02-14
[SQL] Set Spark version to 1.3 using staging dependency

shivaram:
Also fix the maven build in the SparkR SQL branch

cc @cafreeman


concretevitamin:
LGTM

PR #167, opened 2015-02-16
Remove partitionBy() in RDD.

sun-rui:
partitionBy() works on pair RDD, should be in pairRDD.R instead of in RDD.R.
Also move UnionRDD() in RDD.R to the end of file (section for binary functions).


shivaram:
Nice catch -- Thanks @sun-rui . Just to clarify, in the existing code we had partitionBy in two files and so we are removing it from one ?

sun-rui:
yes.

shivaram:
LGTM. Merging this

PR #169, opened 2015-02-16
Sparkr sql

cafreeman:
Merging current master.


shivaram:
@cafreeman -- I think its better to wait for Spark 1.3 to be released before merging this. There are users who run build and run SparkR from source and it'll affect them if they run Spark 1.2 or 1.1 on their clusters.

cafreeman:
@shivaram Actually, this is totally my mistake. I meant to merge all of this into the sparkr-sql branch, not master. Was just trying to update the sparkr-sql branch to reflect the master.

PR #172, opened 2015-02-17
[SPARKR-187] DataFrame to RRDD conversion

cafreeman:
This PR extends the initial work done on the Spark 1.3 DataFrame API by  providing an underlying architecture for converting Spark DataFrames to RRDDs as well as to R data.frames. The conversion to RRDD enables the use of UDFs in lapply (and lapply-like) functions on DataFrames. RRDDs converted from Data Frames will also maintain column names, allowing for UDFs that reference specific columns in the DataFrame as well as the ability to create new columns within the UDF using R's built-in $ operator.

Due to the new serialization format, the "serialized" attribute of the RRDD had to be changed from a boolean to a string. This has broad consequences for the rest of the SparkR codebase due to there being a large number of control flow statements that rely on serialized being boolean.  As a result, RDD's class and constructor definitions have been updated and the underlying utils.R, worker.R, and RRDD.scala components have been updated to reflect this change as well.

In addition, this PR includes the ability to convert a Spark DataFrame to an R data.frame via the `collectToDF()` function. `collectToDF()` calls its own set of Scala functions and re-orients the data in a DataFrame so that it's organized by column instead of row. This enables R to deserialize the data in column-major order (as opposed to Spark's row-major order) and facilitates much faster loading times.


pkg/R/RDD.R
```diff
@@ -1063,7 +1078,7 @@ setMethod("saveAsObjectFile",
           function(rdd, path) {
             # If the RDD is in string format, need to serialize it before saving it because when
             # objectFile() is invoked to load the saved file, only serialized format is assumed.
-            if (!rdd@env$serialized) {
+            if (rdd@env$serialized == "string") {
```
shivaram: Actually this needs to be `rdd@env$serialized != "byte"` because even if the data is serialized as rows we want to convert it to `byte` before saving -- This is because the load code assumes `byte` serialization.

pkg/R/RDD.R
```diff
@@ -1348,16 +1363,16 @@ setMethod("unionRDD",
           function(x, y) {
             if (x@env$serialized == y@env$serialized) {
               jrdd <- callJMethod(getJRDD(x), "union", getJRDD(y))
-              union.rdd <- RDD(jrdd, x@env$serialized)
+              union.rdd <- RDD(jrdd, x@env$serialized, colNames = x@colNames)
             } else {
               # One of the RDDs is not serialized, we need to serialize it first.
-              if (!x@env$serialized) {
+              if (x@env$serialized == "string") {
```
shivaram: Same here, to run the union correctly I think we need to do `!= byte` ? It'll be good to add a testcase which takes 2 data frames, calls `toRDD` on them and then calls `union` on the results to see if this works.

pkg/R/deserialize.R
```diff
@@ -101,3 +101,48 @@ readDeserialize <- function(con) {
...
 
+readDeserializeRows <- function(inputCon) {
+  # readDeserializeRows will deserialize a DataOutputStream composed of
+  # a list of lists. Since the DOS is one continuous stream and
+  # the number of rows varies, we put the readRow function in a while loop
+  # that termintates when the next row is empty.
+  colNames <- readList(inputCon)
+
+  data <- list()
+    numCols <- readInt(inputCon)
```
shivaram: the next 5 lines can be aligned the same as `data <- list()`

pkg/R/utils.R
```diff
@@ -3,56 +3,68 @@
...
-convertJListToRList <- function(jList, flatten, logicalUpperBound = NULL, serialized = TRUE) {
+convertJListToRList <- function(jList, flatten, logicalUpperBound = NULL,
+  serialized = "byte", colNames = list()) {
   arrSize <- callJMethod(jList, "size")
 
   # Unserialized datasets (such as an RDD directly generated by textFile()):
   # each partition is not dense-packed into one Array[Byte], and `arrSize`
   # here corresponds to number of logical elements. Thus we can prune here.
-  if (!serialized && !is.null(logicalUpperBound)) {
+  if (serialized == "string" && !is.null(logicalUpperBound)) {
```
shivaram: Minor style nit: There seems to be two extra spaces at the end of this line

pkg/R/utils.R
```diff
@@ -3,56 +3,68 @@
...
-           })
+            } else {
+              if (inherits(obj, "raw")) {
+                if (serialized == "byte") {
+                  # RDD[Array[Byte]]. `obj` is a whole partition.
+                  res <- unserialize(obj)
+                  # For serialized datasets, `obj` (and `rRaw`) here corresponds to
+                  # one whole partition dense-packed together. We deserialize the
+                  # whole partition first, then cap the number of elements to be returned.
+                  } else {
```
shivaram: Style nit: The changes to this file seem to have some code formatting issue -- The `} else {` here needs to have two less spaces in front (i.e. align with the `if`)

pkg/R/utils.R
```diff
@@ -164,11 +176,11 @@ reserialize <- function(rdd) {
   if (!inherits(rdd, "RDD")) {
     stop("Argument 'rdd' is not an RDD type.")
   }
-  if (rdd@env$serialized) {
-    return(rdd)
-  } else {
+  if (rdd@env$serialized == "string") {
```
shivaram: I think we should do this for `!= "byte"` to be very safe -- I maybe wrong though.
@sun-rui @hlin09  -- Do you have a better idea of whether `reserialize` is required for the new Row-based serialization ?

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SQLUtils.scala
```diff
@@ -1,10 +1,74 @@
...
+
+    SerDe.writeInt(dos, row.length)
+    (0 until row.length).map { idx =>
+      val obj: Object = row(idx).asInstanceOf[Object]
+      SerDe.writeObject(dos, obj)
+    }
+    bos.toByteArray()
+  }
+
+// We convert the DataFrame into an array of RDDs one for each column
```
shivaram: Comments should also be 2 spaces aligned

pkg/R/RDD.R
```diff
@@ -23,7 +24,10 @@ setClass("PipelinedRDD",
 
 
 setMethod("initialize", "RDD", function(.Object, jrdd, serialized,
```
shivaram: At the top of this method we should add a comment saying what the accepted values are for `serialized` and what they mean

pkg/R/RDD.R
```diff
@@ -23,7 +24,10 @@ setClass("PipelinedRDD",
...
-                                        isCached, isCheckpointed) {
+                                        isCached, isCheckpointed, colNames) {
+  # Check that RDD constructor is using the correct version of serialized
+  stopifnot(class(serialized) == "character")
```
shivaram: In addition to this check if will be good to check if `serialized` is equal to either `rows` or `bytes` or `string` just to handle cases where somebody might call it `row` instead of `rows` etc.

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/RRDD.scala
```diff
@@ -157,13 +158,21 @@ private abstract class BaseRRDD[T: ClassTag, U: ClassTag](
...
           }
 
-          for (elem <- iter) {
-            if (parentSerialized) {
-              val elemArr = elem.asInstanceOf[Array[Byte]]
-              dataOut.writeInt(elemArr.length)
-              dataOut.write(elemArr, 0, elemArr.length)
-            } else {
-              printOut.println(elem)
+          if (parentSerialized == "rows") {
```
shivaram: Thinking about this, it might be better to add an object in SerDe.scala which lists these serialization modes. something like 
```
object SerializationFormats {
   val ROWS = "rows"
   val BYTES = "bytes"
   val STRING = "string"
}
```

shivaram:
Thanks @cafreeman for sending this out. This is a great change to have !
@concretevitamin Might be good if you could take a look at some of the changes to `serialized` as well.

pkg/R/utils.R
```diff
@@ -3,56 +3,68 @@
...
-             obj <- callJMethod(jList, "get", as.integer(index))
+          function(index) {
+            obj <- callJMethod(jList, "get", as.integer(index))
 
-             # Assume it is either an R object or a Java obj ref.
-             if (inherits(obj, "jobj")) {
-               if (isInstanceOf(obj, "scala.Tuple2")) {
-                 # JavaPairRDD[Array[Byte], Array[Byte]].
+            # Assume it is either an R object or a Java obj ref.
+              if (inherits(obj, "jobj")) {
```
hlin09: Just style nit: Align `if` with the previous line.

pkg/R/RDD.R
cafreeman: I actually have run that exact test case and the union came out fine. I think this is because PipelinedRDD doesn't draw a distinction between `byte` and `rows` serialization when it's creating the RRDD.

That said, adding this case as an actual unit test is a good idea.  I'll add it now.

pkg/R/utils.R
cafreeman: Turns out the formatting on this function was just a mess after the merge. I've incorporated your change but also tried to straighten out the formatting across the whole thing, so hopefully it'll look better overall.

pkg/R/utils.R
cafreeman: Fixed.

pkg/R/utils.R
cafreeman: Updating the formatting for the whole function since it was all messed up. It'd be great if @hlin09 and @shivaram could give it a second look.

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/RRDD.scala
cafreeman: Would this just be for reference?

shivaram: This would mean that here we could write something like `parentSerialized == SerializationFormats.ROWS` instead of hard-coding strings. In general its better to keep hard coded strings in one file / class and then refer to those variables elsewhere. Makes it easier to extend or modify things.

pkg/R/RDD.R
cafreeman: Added a test case with this scenario.

pkg/R/deserialize.R
cafreeman: Fixed.

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SQLUtils.scala
cafreeman: Fixed.

pkg/R/RDD.R
cafreeman: Added a second check for the three categories.

pkg/R/RDD.R
cafreeman: I've tried it both ways and it doesn't look like simply calling `reserialize` on a "row"-serialized RDD will do the trick. Since all `reserialize` does is recreate the PipelinedRDD via an lapply, I don't think the "row" RDD actually changes at all when you run try to reserialize it.

cafreeman: So I think the real question is how can we programmatically tell `ObjectFile` whether to use `byte` or `row` if the file itself is already serialized? I'm considering putting the `byte` RDD constructor inside a `try` function and then calling `RDD(jrdd, "row")` if it fails, but that feels sort of clumsy.

pkg/R/DataFrame.R
```diff
@@ -114,20 +114,189 @@ setMethod("count",
...
 
-# TODO: Collect() currently returns a list of Generic Row objects and is WIP.  This will eventually 
-# be part of the process to read a DataFrame into R and create a data.frame.
 setMethod("collect",
           signature(rdd = "DataFrame"),
-          function(rdd){
-            sdf <- rdd@sdf
-            listObj <- callJMethod(sdf, "collect")
+          function(rdd) {
+            rddIn <- toRDD(rdd)
```
concretevitamin: Looking at this, can we name the argument to `df` or `sdf`? It's just `toRDD(rdd)` doesn't make sense.

pkg/R/RDD.R
```diff
@@ -13,7 +13,8 @@ setOldClass("jobj")
 #' @export
 setClass("RDD",
          slots = list(env = "environment",
-                      jrdd = "jobj"))
+                      jrdd = "jobj",
+                      colNames = "list"))
```
concretevitamin: Would be good to add a `@slot` comment on the new field.

pkg/R/DataFrame.R
cafreeman: I agree. The only reason it's `rdd` now is because the RDD version of `collect` defines the generic with an argument called `rdd` specifically. This is actually going to likely be an ongoing problem if we continue to have overlap between DataFrame and RDD methods so it'd be good if we figured out a good method for handling this problem in general.

concretevitamin:
@cafreeman 

Just starting a first pass - one general comment about naming: in all the R code, can we rename `serialized` to something like `serializedFormat` or `serializedMode`?

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SerDe.scala
```diff
@@ -282,3 +282,9 @@ object SerDe {
     value.foreach(v => writeBytes(out, v))
   }
 }
+
+object SerializationFormats {
```
concretevitamin: More idiomatic:
```
private sealed trait SerializationFormats
case object ByteSerialized extends SerializationFormats
case object StringSerialized extends SerializationFormats
case object RowSerialized extends SerializationFormats
```

cafreeman:
@concretevitamin Definitely. I had it as `serMode` for a while but changed it back to `serialized` for fear of just changing too many fundamental things, but I think it's a misleading name now that it isn't a boolean.

I know you're just now making a first pass, but be sure to take a look at the comments above about the issues with the new serialization mode and things like `UnionRDD` and `reserialize` with regard to Object Files. Could definitely use your input.

Thanks!

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SQLUtils.scala
```diff
@@ -1,10 +1,74 @@
...
+
+  def getColNames(df: DataFrame): Array[String] = {
+    df.schema.fields.map(_.name)
+  }
+
+  def dfToRowRDD(df: DataFrame): JavaRDD[Array[Byte]] = {
+    df.map(r => rowToRBytes(r))
+  }
+
+  def rowToRBytes(row: Row): Array[Byte] = {
```
concretevitamin: nit: `private[this]` this

pkg/R/utils.R
```diff
@@ -3,56 +3,68 @@
...
+              }
+            } else {
+              if (inherits(obj, "raw")) {
+                if (serialized == "byte") {
+                  # RDD[Array[Byte]]. `obj` is a whole partition.
+                  res <- unserialize(obj)
+                  # For serialized datasets, `obj` (and `rRaw`) here corresponds to
+                  # one whole partition dense-packed together. We deserialize the
+                  # whole partition first, then cap the number of elements to be returned.
+                } else {
```
concretevitamin: Better to be more explicit: `} else if (serialized == "row") { ...`

pkg/inst/tests/test_sparkSQL.R
```diff
@@ -34,4 +34,65 @@ test_that("table() returns a new DataFrame", {
...
+  unioned <- unionRDD(RDD1, RDD2)
+  expect_true(inherits(unioned, "RDD"))
+  expect_true(unioned@env$serialized == "row")
+  expect_true(collect(unioned)[[2]]$name == "Andy")
+})
+
+test_that("lapply() on a DataFrame returns an RDD with the correct columns", {
+  df <- jsonFile(sqlCtx, jsonPath)
+  testRDD <- lapply(df, function(row) {
+    row$newCol <- row$age + 5
```
concretevitamin: This is cool!

pkg/inst/tests/test_sparkSQL.R
```diff
@@ -34,4 +34,65 @@ test_that("table() returns a new DataFrame", {
...
+    row
+    })
+  expect_true(inherits(testRDD, "RDD"))
+  collected <- collect(testRDD)
```
concretevitamin: Just to make sure I understand correctly: is `collected` a list of named lists?

pkg/R/RDD.R
concretevitamin: @cafreeman looking at the [newly added test](https://github.com/cafreeman/SparkR-pkg/commit/8c3b8c5e022e7f747188613bebf5c2b2faedb939), it seems to me the test doesn't hit this else branch, since the two RRDD's serialization modes are equal (RowSerialized).

pkg/inst/tests/test_sparkSQL.R
cafreeman: Yes it is.

pkg/R/RDD.R
cafreeman: Done.

pkg/R/RDD.R
cafreeman: @shivaram @concretevitamin So I think the `serializedMode` problem for `UnionRDD` (and possibly `ObjectFile`) stems from the fact that `reserialize` simply calls `lapply` on the RDD, which then makes use of the fact that getJRDDs default mode is `byte`, thereby causing `getJRDD` to serialize a stringRRDD.

The issue is that `getJRDD` is now updated to dynamically alternate between `row` and `byte` by pulling serialized mode from the incoming RDD itself. This works great when you're only dealing with `row` RRDDs, but it doesn't give an easy way to force the conversion to `byte`.

cafreeman: Pretty sure I figured out how to fix `reserialize` to account for this problem.  I'll be committing the changes soon.

pkg/R/utils.R
hlin09: I lean to agree with @shivaram. Even though current version of reserialize does not actually convert the `row` type, it might be useful to include Row if it does not break other tests.

cafreeman: Agreed. I've updated `reserialize` to account for this, like so:

```R
reserialize <- function(rdd) {
  if (!inherits(rdd, "RDD")) {
    stop("Argument 'rdd' is not an RDD type.")
  }
  if (rdd@env$serializedMode == "string") {
    ser.rdd <- lapply(rdd, function(x) { x })
    return(ser.rdd)
  } else if (rdd@env$serializedMode == "row") {
    rdd@env$serializedMode == "byte"
    ser.rdd <- lapply(rdd, function(x) { x })
    return (ser.rdd)
  } else {
    return(rdd)
  }
}
```
This way, things like `UnionRDD` can handle mixed serialization types and will even return the column names in the `row` serialized part of the union.

pkg/R/utils.R
cafreeman: Fixed.

pkg/inst/tests/test_sparkSQL.R
cafreeman: :) I think so too. Pretty excited about it.

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SerDe.scala
cafreeman: @concretevitamin Not sure I follow (sealed traits are new to me). Would this be in place of the entire object? 
And how would I then go about referencing this in `RRDD.scala` to get the different string values?

pkg/R/DataFrame.R
shivaram: Yeah we can rename the RDD method arguments to be something like `X = "RDD"` -- Lets do this in a separate PR though

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SerDe.scala
shivaram: Yeah I don't sealed traits will do be sufficient here as we need the string value to be passed to R. Sealed traits would be good to use if all usages are within Scala

cafreeman:
@shivaram @concretevitamin Just committed a bunch of changes for the new serialization type. Reworked the `reserialize` function to handle the row type and also changed the name to `serializedMode`. Updated tests accordingly.

At this point I think I've addressed all of the open issues except for the `private[this]` note from @concretevitamin. Wasn't sure what you were asking. Do you just want all the function definitions to be private?

pkg/R/DataFrame.R
```diff
@@ -114,20 +114,189 @@ setMethod("count",
...
+setGeneric("collectToDF", function(df) { standardGeneric("collectToDF") })
+
+setMethod("collectToDF",
+          signature(df = "DataFrame"),
+          function(df) {
+            listCols <- callJStatic("edu.berkeley.cs.amplab.sparkr.SQLUtils", "dfToColRDD", df@sdf)
+            cols <- lapply(seq_along(listCols),
+                           function(colIdx) {
+                             rddCol <- listCols[[colIdx]]
+                             rddRaw <- callJMethod(rddCol, "collect") # Returns a list of byte arrays per partition
```
shivaram: Style comment: I think we use max line length of 100 chars and some of these lines seem longer than that. Can you move the comments to the line above the code to see if it will fix things ?

pkg/R/utils.R
```diff
@@ -164,11 +176,14 @@ reserialize <- function(rdd) {
   if (!inherits(rdd, "RDD")) {
     stop("Argument 'rdd' is not an RDD type.")
   }
-  if (rdd@env$serialized) {
-    return(rdd)
-  } else {
+  if (rdd@env$serializedMode != "byte") {
+    if (rdd@env$serializedMode == "row") {
+      rdd@env$serialzedMode == "byte"
```
shivaram: I am not sure we should be resetting the serialization mode here. During pipelining don't we automatically set the serialization at the end of `getJRDD` to `byte` ?

cafreeman: No, we don't. The default mode in `getJRDD` is `byte`, but that only gets used in deciding whether `StringRRDD` or `RRDD` gets created. Once the function has diverted away from string, the `byte` and `row` versions are the same since we can just pass `colNames` as an empty list and create a normal RRDD. As a result, the serialization type in `getJRDD` is actually defined by the incoming RDD to allow for `worker.R` to correctly differentiate between `byte` and `row`.

Example: https://github.com/cafreeman/SparkR-pkg/blob/sparkr-sql/pkg/R/RDD.R#L159

pkg/R/DataFrame.R
cafreeman: Fixed.

pkg/R/utils.R
shivaram: Right - so the input serialization in `getJRDD` depends on the RDD we pass in. But the output serialization mode is only either `byte` or `string` ? i.e. what I mean is that the serialization mode for the result of `lapply` will be `byte` (even if input serialization mode is `rows` ?)

cafreeman: Only in the case that you are combining a `row` RRDD with another serialization type (e.g. in `UnionRDD` or if you are writing it out with a function that needs it to be `byte` (e.g. `saveAsObjectFile`).

So:
- `lapply` on a `row` RRDD will return another `row` RRDD
- `UnionRDD` on two `row` RRDDs will return a `row` RRDD
- `UnionRDD` on a `row` RRDD and a `byte` or `string` RRDD will return a `byte` RRDD

The `byte` serialization mode is only forced in cases where we we need to call `reserialize`. Otherwise, `getJRDD` will still respect the incoming serialization type.

shivaram: I'm still a little confused by this. I'm going to clone this branch and try out a couple of tests to see if I can better understand how things work. Also longer term we should clean this up better to establish some invariants (for example we should only make `serializedMode` immutable etc.)

cafreeman: Agreed on the invariants. Not sure on the best way to set hard categories for function arguments in R (as opposed to the soft categories imposed by the checks in RDD's initialize function now).

For digging into the reserialize issue, I'd suggest starting with test_sparksql.R as it has tests for unionRDD and objectFile that should provide a good example of how the serializedMode is handled in different cases.

shivaram: I think I've gotten to the bottom of this. Here is my take, @cafreeman @concretevitamin, please correct me if I am wrong.

The problem is in initializing `PipelinedRDD` where we set `serializedMode` - We need the following diff there 
```
-   # This tracks the serialization mode for jrdd_val
-  .Object@env$serializedMode <- prev@env$serializedMode
+  if (!is.null(jrdd_val)) {
+     # This tracks the serialization mode for jrdd_val
+    .Object@env$serializedMode <- prev@env$serializedMode
+  }
```
as we shouldn't set `serializedMode` if `jrdd_val` is NULL (basically the invariant I see is that in PipelinedRDD serializedMode tracks jrdd_val). 

This results in the need for one more change: 
```
-    .Object@env$prev_serializedMode <- .Object@env$serializedMode
+    .Object@env$prev_serializedMode <- prev@env$serializedMode  
```
where we'll set `prev_serializedMode` using `prev@env$serializedMode`

After this we don't need the change in `reserialize` -- The one thing to note is that we shouldn't always assume rdd@env$serialized exists in unit tests or things like that. Basically we don't know the serialization mode for some RDDs if we don't have a JRDD associated with them.

Let me know if this makes sense !

cafreeman: I don't think I'm following, actually. Since the RDD constructor uses `byte` as the default argument for `serializedMode`, wouldn't it necessarily be the case that any RDD you interact with in SparkR would be guaranteed to have a value for `serializedMode`?  I guess I don't understand how we could have an RDD S4 class object without necessarily having a JRDD backing it.

cafreeman:
Thanks to @shivaram's handiwork, the `reserialize` issue has been handled much more elegantly and all the unit tests are looking good again.

@concretevitamin Any more feedback?

shivaram:
I took another look and overall this LGTM  -- @concretevitamin Friendly ping :) if there are no comments I'll merge this in

concretevitamin:
Sorry @cafreeman - been caught up in some stuff. I will take another pass now!

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SerDe.scala
concretevitamin: The nice things about case objects are that they automagically get a bunch of sane methods for free, including toString:
```
scala> case object ByteSerialized
defined module ByteSerialized

scala> ByteSerialized.toString
res0: String = ByteSerialized
```
So wherever we are doing `something == "row"` now, we could instead be more type safe by doing `something == RowSerialized`, and we don't have to worry about writing out the string literals anymore.

Besides this substitution, we just need to change the strings on the R side to match these new names. Thoughts on whether this sounds worthwhile?

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SQLUtils.scala
concretevitamin: Could you mark this as `private[this] def rowToRBytes(row: Row)...`? It is only used once inside this object.

pkg/R/DataFrame.R
```diff
@@ -114,20 +114,192 @@ setMethod("count",
             callJMethod(sdf, "count")
           })
 
-#' Collect elements of a DataFrame
-#' 
-#' Returns a list of Row objects from a DataFrame
-#' 
-#' @param df A SparkSQL DataFrame
-#' 
+# Collect the rows of a Spark DataFrame and return a named list for each row.
```
concretevitamin: More explicit: `Collects a Spark DataFrame and returns a list of named lists (each of which represents a row).`

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/RRDD.scala
```diff
@@ -16,12 +16,13 @@ private abstract class BaseRRDD[T: ClassTag, U: ClassTag](
     parent: RDD[T],
     numPartitions: Int,
     func: Array[Byte],
-    parentSerialized: Boolean,
+    parentSerialized: String,
```
concretevitamin: Let's rename these `parentSerialized` in the Scala classes to match the R names, e.g. `parentSerializedMode`.

pkg/R/RDD.R
```diff
@@ -64,43 +80,62 @@ setMethod("initialize", "PipelinedRDD", function(.Object, prev, func, jrdd_val)
...
   .Object
 })
 
-
 #' @rdname RDD
 #' @export
 #'
 #' @param jrdd Java object reference to the backing JavaRDD
-#' @param serialized TRUE if the RDD stores data serialized in R
+#' @param serialized "byte" if the RDD stores data serialized in R, "string" if the RDD stores strings,
```
concretevitamin: s/serialized/serializedMode

Also add a `@param colNames`

pkg/R/RDD.R
```diff
@@ -64,43 +80,62 @@ setMethod("initialize", "PipelinedRDD", function(.Object, prev, func, jrdd_val)
...
   new("PipelinedRDD", prev, func, NULL)
 }
 
+# Return the serialization mode for an RDD.
+setGeneric("getSerializedMode", function(rdd, ...) { standardGeneric("getSerializedMode") })
+# For normal RDDs we can directly read the serializedMode
+setMethod("getSerializedMode", signature(rdd = "RDD"), function(rdd) rdd@env$serializedMode )
+# For pipelined RDDs if jrdd_val is set then serializedMode should exist
+# if not we return the defaultSerialization mode of "byte" as we don't know the serialization
+# mode at this point of time.
```
concretevitamin: nit: "at this point in time"

pkg/R/RDD.R
```diff
@@ -64,43 +80,62 @@ setMethod("initialize", "PipelinedRDD", function(.Object, prev, func, jrdd_val)
...
+setMethod("getSerializedMode", signature(rdd = "PipelinedRDD"),
+          function(rdd, defaultSerialization = "byte") {
```
concretevitamin: The argument `defaultSerialization` doesn't seem intended to be supplied by a caller -- makes sense to just move it inside the function body?

pkg/R/RDD.R
```diff
@@ -64,43 +80,62 @@ setMethod("initialize", "PipelinedRDD", function(.Object, prev, func, jrdd_val)
     .Object@prev_jrdd <- getJRDD(prev)
     # Since this is the first step in the pipeline, the prev_serialized
     # is same as serialized here.
-    .Object@env$prev_serialized <- .Object@env$serialized
+    .Object@env$prev_serializedMode <- prev@env$serializedMode
```
concretevitamin: With the new change, these two lines of comments are stale and should be removed

pkg/R/RDD.R
```diff
@@ -46,8 +58,12 @@ setMethod("initialize", "PipelinedRDD", function(.Object, prev, func, jrdd_val)
...
-   # This tracks if jrdd_val is serialized
-  .Object@env$serialized <- prev@env$serialized
+  if (!is.null(jrdd_val)) {
+     # This tracks the serialization mode for jrdd_val
+    .Object@env$serializedMode <- prev@env$serializedMode
+  }
+  
+  .Object@colNames <- prev@colNames
 
   # NOTE: We use prev_serialized to track if prev_jrdd is serialized
```
concretevitamin: Update comment: (1) s/prev_serialized/prev_serializedMode (2) not checking whether it is serialized now -- checking what mode it is serialized in

pkg/R/RDD.R
```diff
@@ -64,43 +80,62 @@ setMethod("initialize", "PipelinedRDD", function(.Object, prev, func, jrdd_val)
...
+              return(defaultSerialization)
+            }
+          })
 
 # The jrdd accessor function.
 setGeneric("getJRDD", function(rdd, ...) { standardGeneric("getJRDD") })
 setMethod("getJRDD", signature(rdd = "RDD"), function(rdd) rdd@jrdd )
 setMethod("getJRDD", signature(rdd = "PipelinedRDD"),
-          function(rdd, dataSerialization = TRUE) {
+          function(rdd, dataSerialization = "byte") {
```
concretevitamin: Would be good to be consistent in naming now that we have a clear grasp of the issues

pkg/R/RDD.R
```diff
@@ -1063,7 +1102,7 @@ setMethod("saveAsObjectFile",
           function(rdd, path) {
             # If the RDD is in string format, need to serialize it before saving it because when
```
concretevitamin: Comment is stale

pkg/R/utils.R
```diff
@@ -3,56 +3,68 @@
 # Given a JList<T>, returns an R list containing the same elements, the number
 # of which is optionally upper bounded by `logicalUpperBound` (by default,
 # return all elements).  Takes care of deserializations and type conversions.
-convertJListToRList <- function(jList, flatten, logicalUpperBound = NULL, serialized = TRUE) {
+convertJListToRList <- function(jList, flatten, logicalUpperBound = NULL,
+  serializedMode = "byte", colNames = list()) {
   arrSize <- callJMethod(jList, "size")
 
   # Unserialized datasets (such as an RDD directly generated by textFile()):
```
concretevitamin: Change "Unserialized datasets" to "Datasets with serializedMode 'string'"

pkg/R/utils.R
```diff
@@ -164,11 +176,11 @@ reserialize <- function(rdd) {
   if (!inherits(rdd, "RDD")) {
```
concretevitamin: Let's update the comment of this function to explicitly say it reserializes to Byte mode.

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SerDe.scala
concretevitamin: BTW regardless of what we decide on the above proposal - let's be consistent in naming and change `SerializationFormats` to the name on R side (`SerializedModes` or `SerializationModes`).

concretevitamin:
I did a pass and I don't think there's anything major left -- code is quite clear, thanks @cafreeman @shivaram !

One more general comment: what do you guys think about renaming `reserialize` to `reserializeToBytes`? Sounds clearer to me.

pkg/R/DataFrame.R
```diff
@@ -114,20 +114,192 @@ setMethod("count",
...
+
+setMethod("collectToDF",
+          signature(df = "DataFrame"),
+          function(df) {
+            listCols <- callJStatic("edu.berkeley.cs.amplab.sparkr.SQLUtils", "dfToColRDD", df@sdf)
+            cols <- lapply(seq_along(listCols),
+                           function(colIdx) {
+                             rddCol <- listCols[[colIdx]]
+                             # Returns a list of byte arrays per partition
+                             rddRaw <- callJMethod(rddCol, "collect")
```
davies: Instead to call collect for each column, I'd like to call collect for df once, then convert the result (list of row) into list of column in the driver (In Scala or R).

pkg/R/DataFrame.R
```diff
@@ -114,20 +114,192 @@ setMethod("count",
...
+#' }
+
+setGeneric("toRDD", function(df) { standardGeneric("toRDD") })
+
+setMethod("toRDD",
+          signature(df = "DataFrame"),
+          function(df) {
+            jrdd <- callJStatic("edu.berkeley.cs.amplab.sparkr.SQLUtils", "dfToRowRDD", df@sdf)
+            names <- callJStatic("edu.berkeley.cs.amplab.sparkr.SQLUtils", "getColNames", df@sdf)
+            RDD(jrdd, serializedMode = "row", colNames = names)
```
davies: I think we can convert the row in jrdd into named list without having colNames in RDD:

```
zip <- function(x, y){
    stopifnot(length(x)==length(y))
    z <- list()
    for (i in seq_along(x)){
        z[[i]] <- list(x[[i]], y[[i]]) 
    }
    z
}
rdd <- RDD(jrdd, serializedMode = "row")
lapply(rdd, function (row) { zip(row, names) })
```

pkg/R/RDD.R
```diff
@@ -21,9 +24,16 @@ setClass("PipelinedRDD",
...
          contains = "RDD")
 
-
-setMethod("initialize", "RDD", function(.Object, jrdd, serialized,
-                                        isCached, isCheckpointed) {
+setMethod("initialize", "RDD", function(.Object, jrdd, serializedMode,
+                                        isCached, isCheckpointed, colNames) {
+  # Check that RDD constructor is using the correct version of serializedMode
+  stopifnot(class(serializedMode) == "character")
+  stopifnot(serializedMode %in% c("byte", "string", "row"))
```
davies: byte -> raw ?

pkg/R/RDD.R
```diff
@@ -46,8 +58,12 @@ setMethod("initialize", "PipelinedRDD", function(.Object, prev, func, jrdd_val)
...
   .Object@env$isCheckpointed <- FALSE
   .Object@env$jrdd_val <- jrdd_val
-   # This tracks if jrdd_val is serialized
-  .Object@env$serialized <- prev@env$serialized
+  if (!is.null(jrdd_val)) {
+     # This tracks the serialization mode for jrdd_val
+    .Object@env$serializedMode <- prev@env$serializedMode
+  }
+  
+  .Object@colNames <- prev@colNames
```
davies: `colNames` can not be inherited from prev, if we can not make sure that `func` does keep them not changed.

pkg/R/RDD.R
davies: Is there any way to mark this method as internal api?

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SerDe.scala
davies: I'd like to continue with String, because it's easy to be passed between Scala and R, or you will have some helpful function to convert the string into Java object.

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SQLUtils.scala
```diff
@@ -1,10 +1,74 @@
...
+import java.io.ByteArrayOutputStream
+import java.io.DataOutputStream
 
 object SQLUtils {
   def createSQLContext(sc: SparkContext): SQLContext = {
     new SQLContext(sc)
   }
+
+  def getColNames(df: DataFrame): Array[String] = {
+    df.schema.fields.map(_.name)
```
davies: df.columns.toArray

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SQLUtils.scala
```diff
@@ -1,10 +1,74 @@
...
+  // Each RDD contains a serialized form of the column per partition.
+  def dfToColRDD(df: DataFrame): Array[RDD[Array[Byte]]] = {
+    val colRDDs = convertRowsToColumns(df)
+    val dfOut = colRDDs.map { col =>
+      colToRBytes(col)
+    }
+    dfOut
+  }
+
+  def convertRowsToColumns(df: DataFrame): Array[RDD[Any]] = {
```
davies: This is not efficient if we have lots of columns.

pkg/R/DataFrame.R
```diff
@@ -114,20 +114,192 @@ setMethod("count",
...
+#'\dontrun{
+#' sc <- sparkR.init()
+#' sqlCtx <- sparkRSQL.init(sc)
+#' path <- "path/to/file.json"
+#' df <- jsonFile(sqlCtx, path)
+#' converted <- collectToDF(df)
+#' is.data.frame(converted)
+#' }
+
+setGeneric("collectToDF", function(df) { standardGeneric("collectToDF") })
```
davies: Is it possible that change `collect` to return an data frame instead of having a new API `collectToDF`, because data frame is so nature that almost all the people will want it.

pkg/R/DataFrame.R
```diff
@@ -114,20 +114,192 @@ setMethod("count",
...
+#' sqlCtx <- sparkRSQL.init(sc)
+#' path <- "path/to/file.json"
+#' df <- jsonFile(sqlCtx, path)
+#' take(df, 2)
+#' }
+
+setMethod("take",
+          signature(rdd = "DataFrame", num = "numeric"),
+          function(rdd, num) {
+            rddIn <- toRDD(rdd)
```
davies: `rdd.limit(num).collect()` will be better, which could benefit from filter push down from DataFrame

davies:
@cafreeman This PR looks pretty cool, I had left some comments, hope that we can simplify this PR a little bit, thanks!

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SerDe.scala
concretevitamin: That's a good point; the easiest way I can think of is one additional implicit on Scala side, which does sound like more code to maintain :)  Feel free to drop this one @cafreeman.

pkg/R/DataFrame.R
shivaram: This is a great idea @davies ! If we can make this to work, it'll simplify the RDD class on both R and Scala side.

pkg/R/DataFrame.R
shivaram: I think @cafreeman measured that it was slower to do the conversion row-wise instead of column-wise on the R-side. However we could do a `collect` of the DF and do the row to column wise conversion in Scala on the driver.

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SQLUtils.scala
cafreeman: Done.

pkg/R/DataFrame.R
cafreeman: Good call. That's better.

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/RRDD.scala
cafreeman: Fixed.

pkg/R/RDD.R
cafreeman: Fixed.

pkg/R/RDD.R
cafreeman: Fixed.

pkg/R/RDD.R
cafreeman: Done.

pkg/R/RDD.R
cafreeman: Changed to `serializedMode`

pkg/R/RDD.R
cafreeman: Updated.

pkg/R/utils.R
cafreeman: Fixed.

pkg/R/utils.R
cafreeman: Fixed.

pkg/R/DataFrame.R
cafreeman: @davies Does the new DataFrame API provide any new methods for collecting a DataFrame differently? The problem here is that, since R stores everything as column-major, having to transpose everything when converting to a `data.frame` is really, really inefficient. That said, I know the new API does have a "Column" class. Is there now a built-in way to re-orient the DataFrame so that it's truly stored as columns instead of rows?

davies: No. The loop in R is slow, but loop in Scala much faster, so I think we could collect all the columns in Scala, then serialize them in column-major way, finally R could get the data.frame in column-major.

The end-to-end performance will be better than the current approach, especially for the multiple columns DataFrames, because of the overhead of schedule a Spark job (maybe compute the RDD multiple times).
The total data must be fit in memory, so total size can not be huge.

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SQLUtils.scala
cafreeman: Any suggestions on a more efficient way? Even with the suggestions above, we still need to transpose the entire DF at some point so that rows become columns.

cafreeman: What if we create an array of names with `df.fields` and then iterate through the names array and call the apply function on the DataFrame for each column name. This would let us grab each column of the DataFrame individually and then we could serialize from there.

shivaram: @cafreeman I think @davies suggestion was related to the comment about each column's `collect` triggering a lot of jobs. So the way to run just one `collect` would be to do something like this in Scala

```
def dfToCols(df: DataFrame) = {
  val localDF = df.collect() # This will be Seq[Row]
  val dfCols = convertRowsToColumns(localDF) # This will Array[Array[Any]] instead of Array[RDD[Any]]
  dfCols.map { col => colToRBytes(col) } # Again colToRBytes will take in Array[Any]
}
```

pkg/R/DataFrame.R
shivaram: I think this is related to the efficiency comment in https://github.com/amplab-extras/SparkR-pkg/pull/172/files#r25185321 -- The change should be pretty small actually (mostly just moving the collect into Scala to be run once).

pkg/R/RDD.R
cafreeman: I prefer `byte` in this case solely because it makes the code a lot clear at a glance. `row` and `raw` look very similar when you're skimming.

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SQLUtils.scala
cafreeman: I just dropped this functions out and called `columns` with `callJMethod` instead. Makes things even simpler.

pkg/R/DataFrame.R
cafreeman: Done.

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SQLUtils.scala
cafreeman: I've reworked `SQLUtils.scala` and the `collect` method for DataFrames so that `collect` now returns a DataFrame and the row to column conversion is (hopefully) much more optimized by calling a single collect.

pkg/R/DataFrame.R
cafreeman: I'm not sure it's quite that easy. The `colNames` slot was actually added because I needed a way to pass the column names into `RRDD.scala` and, subsequently, to the R workers when evaluating UDFs from the R side. By putting them into the `colNames` slot, we ensure they have a placeholder when it comes time to deserialize all of the information on the R worker side after the `RRDD` constructor has been called. This also factors into PipelinedRDDs and being able to perform multiple transformation on an RDD in `row` mode.

I'm not 100% sure either way and I think it's worth discussing more, but I do know that there's more to it than just needing to lapply names onto an RDD after the fact.

What do you guys think?

cafreeman:
Thanks for the comments @davies! I appreciate the feedback and you pointed out a lot of useful things. I've incorporated a lot of your suggestions in the recent commits. Still trying to figure out the best way to handle `colNames` in RRDDs, but the other stuff should be good to go so feel free to take a second look. Thanks!

pkg/R/DataFrame.R
cafreeman: To follow up, I definitely see how mapping the names onto the RRDD would definitely work on some levels. It would mean that `toRDD` would result in a PipelinedRDD right away, that and that PipelinedRDD would apply the column names to every row in the the RRDD when it is calculated. So far so good.

Where I'm still hesitant is how the R workers handle this across multiple transformations. If we want to `lapply` multiple UDFs in sequence and reference existing/create new columns, will the lack of an actual slot in the RRDD become a problem, or do the names just exist as part of the data regardless? @shivaram, we went back and forth on this a bit initially. What's your take?

pkg/R/RDD.R
shivaram: Yeah `byte` seems reasonable for that reason. FWIW we should try to isolate the uses of these strings by writing a enum-like class or something like that in a future PR.

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SerDe.scala
shivaram: That sounds more complicated :) I'd vote to keep it as strings for now

shivaram:
@cafreeman -- Regarding the `colNames` thing, your analysis that we could create a pipelined RDD is correct. These names will become saved as a part of the data and should be retained across transformations. In fact this is possible even with text files today if the user sets up the name. For example, lets say I have a file with entries
```
Andy,30
Justin,19
```

Then we can run SparkR to create an RDD with something like
```
rdd <- textFile(sc, "/tmp/test-file")
listRDD <- lapply(rdd, 
     function(line) { 
        parts = strsplit(line, ",", fixed=T); 
        list(name=parts[[1]], age=as.integer(parts[[2]])) 
     })
take(listRDD, 1L)
newColRDD <- lapply(listRDD, function(li) { li$newAge <- li$age + 1; li })
take(newColRDD, 1L)
```

This happens because the `byte` serialized also serializes the column names and maintains it with every RDD transformation.

pkg/R/DataFrame.R
```diff
@@ -114,20 +114,147 @@ setMethod("count",
             callJMethod(sdf, "count")
           })
 
-#' Collect elements of a DataFrame
-#' 
-#' Returns a list of Row objects from a DataFrame
-#' 
-#' @param df A SparkSQL DataFrame
-#' 
+# Collects all the elements of a Spark DataFrame and coerces them into an R data.frame.
```
davies: #'

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SQLUtils.scala
```diff
@@ -1,10 +1,68 @@
...
+    }
+    bos.toByteArray()
+  }
+
+  def dfToCols(df: DataFrame): Array[Array[Byte]] = {
+    // localDF is Array[Row]
+    val localDF = df.collect()
+    val numCols = df.columns.length
+    // dfCols is Array[Array[Any]]
+    val dfCols = convertRowsToColumns(localDF, numCols)
```
davies: This could be optimized to generate a Array[Byte] directly, it could avoid another pass of serialization for returned Array[Array[Byte]]

pkg/R/DataFrame.R
```diff
@@ -114,20 +114,147 @@ setMethod("count",
...
+            lapplyPartition(rdd, FUN)
+          })
+
+setMethod("mapPartitions",
+          signature(X = "DataFrame", FUN = "function"),
+          function(X, FUN) {
+            lapplyPartition(X, FUN)
+          })
+
+setMethod("lapplyPartitionsWithIndex",
```
davies: We did not have this in Scala/Python

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/RRDD.scala
```diff
@@ -157,13 +158,21 @@ private abstract class BaseRRDD[T: ClassTag, U: ClassTag](
...
-              val elemArr = elem.asInstanceOf[Array[Byte]]
-              dataOut.writeInt(elemArr.length)
-              dataOut.write(elemArr, 0, elemArr.length)
-            } else {
-              printOut.println(elem)
+          if (parentSerializedMode == SerializationFormats.ROW) {
+            SerDe.writeStringArr(dataOut, colNames)
+            for (row <- iter) {
+              val rowArr = row.asInstanceOf[Array[Byte]]
+              dataOut.write(rowArr, 0, rowArr.length)
```
davies: nit: dataOut.write(rowArr)

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/RRDD.scala
```diff
@@ -157,13 +158,21 @@ private abstract class BaseRRDD[T: ClassTag, U: ClassTag](
...
+            }
+          } else {
+            for (elem <- iter) {
+              if (parentSerializedMode == SerializationFormats.BYTE) {
+                val elemArr = elem.asInstanceOf[Array[Byte]]
+                dataOut.writeInt(elemArr.length)
+                dataOut.write(elemArr, 0, elemArr.length)
+              } else {
+                printOut.println(elem)
```
davies: add a TODO here: handle utf-8 encoding

pkg/man/sparkRSQL.init.Rd
```diff
@@ -1,7 +1,7 @@
 % Generated by roxygen2 (4.0.2): do not edit by hand
 \name{sparkRSQL.init}
 \alias{sparkRSQL.init}
-\title{Initialize a new Spark Context.}
+\title{Initialize a new SQLContext.}
```
davies: These are generated docs, should we keep them out of git?

davies:
For the discussion about colNames, I think both of them has similar runtime behavior in cases of UDFs:

```
rdd <- toRDD(df)
rdd1 <- lapply(rdd, udf1)
rdd2 <- lapply(rdd1, udf2)
```

Because we do not know the behavior inside `udf1`, so `rdd1` will be a general RDD without `colNames` (PipelinedRDD actually). `rdd2` also could not have colNames. Each row in rdd1 and rdd2 will be named list.

In the worker, when it calculate rdd2, it will readSerializedRow() from file, call assignNames() on them, then call udf1, and udf2:   list -> named list -> udf1() -> udf2(). Both of the two approach, will do the same thing.

The difference is that, when you call `collect(toRDD(df))`, The current approach will call `assignNames` on driver, but the new approach will call assignNames on workers, will having more data to serialize (including names for each Row). But this case is not common, most of the user will call `collect(df)` or `lapply(df, udf)`.

pkg/R/RDD.R
```diff
@@ -62,45 +76,65 @@ setMethod("initialize", "PipelinedRDD", function(.Object, prev, func, jrdd_val)
...
     .Object@func <- func
     .Object@prev_jrdd <- getJRDD(prev)
-    # Since this is the first step in the pipeline, the prev_serialized
-    # is same as serialized here.
-    .Object@env$prev_serialized <- .Object@env$serialized
+    .Object@env$prev_serializedMode <- prev@env$serializedMode
+    # NOTE: We use prev_serializedMode to track the serialization mode of prev_JRDD
+    # prev_serializedMode is used during the delayed computation of JRDD in getJRDD
+
+    .Object@colNames <- prev@colNames
```
davies: This is not correct, `func` can do anything (including change the number columns, or column name), so PipelinedRDD@colNames should be `list()`.

Only in few of cases, `.Object@colNames` can be `prev@colNames`, for example, reserialization.

pkg/R/DataFrame.R
shivaram: Hmm - Is there any specific reason it wasn't added ? It seems like something useful to have.

pkg/man/sparkRSQL.init.Rd
shivaram: Yeah, so far we have kept the generated docs in git as we don't have a nice way to get it into the installed package. We can revisit this when we try to merge with mainline Spark.

pkg/R/DataFrame.R
davies: I think this is the most fundamental API for RDD, but is not used frequently in practice. We can easily convert an DataFrame into a RDD, so these short-hand APIs is not that necessary, I'd like to keep them as few as possible. 

cc @rxin

cafreeman:
At this point it seems like dropping all references to `colNames` is the correct move. By lapplying the names in the toRDD function, @davies was right, we no longer need the slot at all. As a result, I've removed the colNames references from everything, including the Scala side as well as `worker.R`.

cafreeman:
The only downside to this approach is that `toRDD` now returns a PipelinedRDD which throws off @shivaram's `getSerializedMode` function since `jrdd_val` defaults to `NULL` we end up always getting `serializedMode == byte`

davies:
@cafreeman That means we do not need `row` mode anymore, is't cool?

pkg/R/DataFrame.R
cafreeman: I've removed with `withIndex` functions. The remaining functions mirror what is currently in PySpark.

cafreeman:
I've implemented all the changes discussed in our conversation earlier today. The `colNames` slot is now a thing of the past and everything has been updated accordingly.

I also added a `limit` function to the `DataFrame` class and reworked `take` so that it would make use of the catalyst optimizer instead of converting to an RDD and using the RDD version of `take`.

At this point, all tests pass and I believe we should be good to go unless anyone else has other issues.

Thanks again for all the feedback @shivaram @concretevitamin @davies

davies:
This looks good to me now, thanks!

shivaram:
Thanks @cafreeman for working through the comments. Merging this.

PR #173, opened 2015-02-18
[SPARKR-200][SPARKR-149] Fix path, classpath separator for Windows

shivaram:
This PR fixes a couple of problems reported by users on Windows

- The first fix is to make the classpath separator `;` as described in [1]
- The second fix is to handle spaces in the install path (in Windows, Unix-like platforms). The solution that seems to work across all platforms is to convert the jar paths to URIs and encode the URIs using `URLEncode`

Tested on Mac, Windows VM and Linux VM. P.S: Reviewers might find it useful to test using something like
```
library(devtools)
with_lib("/tmp/sparkr install", install_github(repo = "shivaram/SparkR-pkg/pkg@windows-space-fix"))
```

[1] http://stackoverflow.com/a/4528456/4577954

### DIFF
--- a/pkg/R/sparkR.R
+++ b/pkg/R/sparkR.R
@@ -90,10 +90,13 @@ sparkR.init <- function(
     normalizePath(c(as.character(.sparkREnv$assemblyJarPath), as.character(sparkJars))))
 
   # Classpath separator is ";" on Windows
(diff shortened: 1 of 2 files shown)


felixcheung:
LGTM

shivaram:
Thanks @felixcheung for taking a look. Merging this.

PR #174, opened 2015-02-18
[SPARKR-178] Integrate with SparkR with spark-submit

shivaram:
This patch integrates SparkR with spark-submit and allows SparkR to be used in `yarn-client` and `yarn-cluster` mode. This is done by adding a new SparkRRunner class which is invoked as the `main` class by `spark-submit`.  As the JVM is launched using spark-submit we automatically pick up Spark config variables in this case.

For the non-yarn use cases, this patch also handles reading in the default Spark config variables (in spark-defaults.conf or passed through cmdline) by launching the SparkRBackend through spark-submit.

Tested on standalone mode on Mac and on a YARN cluster in Cloudera VM. I'll try to test this on a EC2 cluster with YARN soon.

cc @felixcheung -- It'll be great if you can help test this version of the patch on your YARN cluster !


felixcheung:
I've updated everything to your PR and it gets an error:

```
$ export YARN_CONF_DIR=/etc/hadoop/conf
$ export SPARK_HOME=/opt/cloudera/parcels/CDH-5.3.0-1.cdh5.3.0.p0.30/lib/spark
$ export JAVA_HOME=/usr/java/jdk1.7.0_67-cloudera
$ /usr/lib64/R/library/SparkR/sparkR-submit --master yarn-client ./pi.R yarn-client 4
Running /opt/cloudera/parcels/CDH-5.3.0-1.cdh5.3.0.p0.30/lib/spark/bin/spark-submit --class edu.berkeley.cs.amplab.sparkr.SparkRRunner --files ./pi.R --master yarn-client /usr/lib64/R/library/SparkR/sparkr-assembly-0.1.jar ./pi.R yarn-client 4
Loading required package: methods
[SparkR] Initializing with classpath /usr/lib64/R/library/SparkR/sparkr-assembly-0.1.jar

Error in socketConnection(host = hostname, port = port, server = FALSE,  :
  cannot open the connection
Calls: sparkR.init -> connectBackend -> socketConnection
In addition: Warning message:
In socketConnection(host = hostname, port = port, server = FALSE,  :
  localhost:12345 cannot be opened
Execution halted
```

felixcheung:
And it works when I run R

shivaram:
Thanks @felixcheung  for testing -- can you check if the SparkRBackend comes up successfully in the case of running `pi.R` ? You can do this with `jps` or something like that. I am wondering if the problem is that it takes longer than the time we wait (2s) for the process to come up or if the process dies for some other reason.

felixcheung:
It's possible. I just rerun sparkR-submit again after successfully running the pi.R code in R, and sparkR-submit works.
AFAIK, we are not caching/keeping the JVM process here, and so any idea why it works the 2nd time but not the 1st?

shivaram:
I am not sure there is a deterministic reason. I think it just happens that sometimes it takes longer to open ports etc. There is a open PR to retry these connection requests instead of failing after trying once https://sparkr.atlassian.net/browse/SPARKR-193

shivaram:
Specifically in the SparkRRunner case we have two threads which are kind of racing -- The first thread starts the backend and the second thread starts the R process. Actually I think I can actually add a synchronization here. Let me do that and ping this PR again

felixcheung:
Great!
oh and perhaps it should use a random port to avoid conflict (eg. multiple users or jobs), like pyspark/PythonRunner does.

shivaram:
Thats a good point too - Just pushed out a change with the synchronization and random port assignment.

felixcheung:
Retested this, a few times, everything is working well!

One side issue is I've noticed a hardcoded port when running sparkR.init in R directly:
sc <- sparkR.init(Sys.getenv("MASTER", unset = "local"))
Launching java with command  /usr/java/jdk1.7.0_67-cloudera/bin/java   -Xmx512m -cp /usr/lib64/R/library/SparkR/sparkr-assembly-0.1.jar::/etc/hadoop/conf edu.berkeley.cs.amplab.sparkr.SparkRBackend 12345
SparkR Backend server started on port :12345

shivaram:
The reason for this is mostly because we exposed `sparkRBackendPort` as an option to `sparkR.init` with a default value of 12345. 

Thinking more about this I am worried that we will have collisions with other Spark or Hadoop services (for example at 7077 or 8080 or 50030 etc). I think the right thing to do is to start at a well known port and then incrementally check for available ports.  However I am worried this adds too much complexity to this PR and I'm thinking it'll be better to use a fixed port for now and track the random port with a new PR.

shivaram:
Note that we will have a config variable `SPARKR_BACKEND_PORT` that users can set in case they want to run multiple backends on the same machine etc.

felixcheung:
That's fair. I think Py4J (in pyspark) starts with a default port and increments if there is a conflict.
http://py4j.sourceforge.net/getting_started.html

shivaram:
@concretevitamin - Could you take a look at this ? I think this has been tested by me and @felixcheung but it'll be good to get another pair of eyes

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRRunner.scala
```diff
@@ -0,0 +1,97 @@
...
+      rFile
+    }
+
+
+    // Launch a SparkR backend server for the R process to connect to; this will let it see our
+    // Java system properties etc.
+    val sparkRBackend = new SparkRBackend()
+    val sparkRBackendThread = new Thread() {
+      val finishedInit = new Semaphore(1)
+      finishedInit.acquire()
```
concretevitamin: It seems to me we should

* initialize the semaphore to have 0 permit
* remove the acquire inside here (but keep the one in the caller thread)

The current setup seems to have a risk of the caller thread acquiring first, then this constructor will deadlock?

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRRunner.scala
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    sparkRBackendThread.start()
+    // Wait for SparkRBackend initialization to finish
+    sparkRBackendThread.finishedInit.acquire()
+
+    // Launch R
+    val builder = new ProcessBuilder(Seq(rCommand, rFileNormalized) ++ otherArgs)
+    val env = builder.environment()
+    env.put("EXISTING_SPARKR_BACKEND_PORT", "" + sparkRBackendPort)
```
concretevitamin: nit: `.toString` is more direct

pkg/R/sparkRClient.R
```diff
@@ -39,3 +39,30 @@ launchBackend <- function(
...
 }
+
+launchBackendSparkSubmit <- function(
+    mainClass,
+    args,
+    appJar,
+    sparkHome,
+    sparkSubmitOpts) {
+  if (.Platform$OS.type == "unix") {
+    spark_submit_bin_name = "spark-submit"
```
concretevitamin: style nit: camelCase here and in the if block below

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRRunner.scala
shivaram: fixed

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRRunner.scala
shivaram: Well the current setup did work because the constructor of the thread is run completely before the object is returned. So both `new Semaphore` and `acquire` will run before the object is returned. 

But I like your suggestion of initializing semaphore is `0` -- I've changed it to do that.

pkg/R/sparkRClient.R
shivaram: done

pkg/R/sparkR.R
```diff
@@ -106,11 +106,28 @@ sparkR.init <- function(
...
+
+  sparkRExistingPort <- Sys.getenv("EXISTING_SPARKR_BACKEND_PORT", "")
+  if (sparkRExistingPort != "") {
+    sparkRBackendPort <- sparkRExistingPort
+  } else {
+    if (Sys.getenv("SPARKR_USE_SPARK_SUBMIT", "") == "") {
+      launchBackend(classPath = cp,
+                    mainClass = "edu.berkeley.cs.amplab.sparkr.SparkRBackend",
+                    args = as.character(sparkRBackendPort),
+                    javaOpts = paste("-Xmx", sparkMem, sep = ""))
```
felixcheung: Should this be reconciled with spark.executor.memory or spark.driver.memory?

pkg/inst/sparkR-submit
```diff
@@ -0,0 +1,60 @@
...
+
+  primary="${APPLICATION_OPTS[0]}"
+  shift
+  # Set the main class to SparkRRunner and add the primary R file to --files to make sure its copied to the cluster
+  echo "Running $SPARK_HOME/bin/spark-submit --class edu.berkeley.cs.amplab.sparkr.SparkRRunner --files $primary ${SUBMISSION_OPTS[@]} $SPARKR_JAR_FILE $primary" "${APPLICATION_OPTS[@]:1}"
+  exec "$SPARK_HOME"/bin/spark-submit --class edu.berkeley.cs.amplab.sparkr.SparkRRunner --files "$primary" "${SUBMISSION_OPTS[@]}" "$SPARKR_JAR_FILE" "$primary" "${APPLICATION_OPTS[@]:1}"
+else
+
+  # If we don't have an R file to run, run R shell
+  R
```
felixcheung: Should it create the spark context here?

pkg/R/sparkR.R
shivaram: This should be the same as `spark.driver.memory`. However we can't get that value here as the conf is parsed by `spark-submit`. However I see this code path as something we have mostly for backwards compatibility -- Going forward we will deprecate the use case of not using spark-submit

pkg/inst/sparkR-submit
shivaram: I thought this would be tricky as we don't know the lib install dir, but the library is always available at `../` relative to this script as this script is in the `inst

pkg/inst/sparkR-submit
```diff
@@ -0,0 +1,74 @@
...
+.First <- function() {
+  projecHome <- Sys.getenv("PROJECT_HOME")
+  Sys.setenv(NOAWT=1)
+  .libPaths(c(paste(projecHome,"/..", sep=""), .libPaths()))
+  require(SparkR)
+  sc <- sparkR.init()
+  assign("sc", sc, envir=.GlobalEnv)
+  cat("\n Welcome to SparkR!")
+  cat("\n Spark context is available as sc\n")
+}
```
felixcheung: Actually when R deploy/install this, it put the sparkR-submit script at the root of the library package, ie. same as the sparkr-assembly.jar, and not under inst.

So I think you don't need to put `.libPaths(c(paste(projecHome,"/..", sep=""), .libPaths()))`

This is also why I change SPARKR_JAR_FILE earlier.

shivaram: @felixcheung That works if the install path is standard (for example we call `R CMD INSTALL pkg`)  -- But if the installation happens to a custom dir (say in my case I installed it to `/home/shivaram/SparkR-pkg/lib` then library(SparkR) doesn't work.

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRRunner.scala
```diff
@@ -0,0 +1,96 @@
...
+    val builder = new ProcessBuilder(Seq(rCommand, rFileNormalized) ++ otherArgs)
+    val env = builder.environment()
+    env.put("EXISTING_SPARKR_BACKEND_PORT", sparkRBackendPort.toString)
+    builder.redirectErrorStream(true) // Ugly but needed for stdout and stderr to synchronize
+    val process = builder.start()
+
+    new RedirectThread(process.getInputStream, System.out, "redirect output").start()
+
+    val returnCode = process.waitFor()
+    sparkRBackendThread.stopBackend()
```
felixcheung:
```
val returnCode = try {
 val process = builder.start()
 new RedirectThread(process.getInputStream, System.out, "redirect output").start()
 process.waitFor()
} finally {
  sparkRBackendThread.stopBackend()
}
System.exit(returnCode)
```

This is a bit hard to read, but I think we should put stopBackend in a finally in case there is an error with builder.start().
Note that we can't have the System.exit inside the try because then the finally will not execute.
http://www.mytechnotes.biz/2012/07/about-using-systemexit-in-try-catch.html

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRRunner.scala
```diff
@@ -0,0 +1,96 @@
...
+      }
+
+      def stopBackend() {
+        sparkRBackend.close()
+      }
+    }
+
+    sparkRBackendThread.start()
+    // Wait for SparkRBackend initialization to finish
+    sparkRBackendThread.finishedInit.acquire()
```
felixcheung: Should call `tryAquire(long timeout, TimeUnit unit)`? In case sparkRBackend.init throws and it never releases the semaphore, this could wait forever.

pkg/inst/sparkR-submit
felixcheung: Good to know, maybe it should check what's in `/..` then?
For instance, `R CMD INSTALL pkg` would put sparkR-submit at /usr/lib64/R/library/SparkR, would adding /.. of that which is /usr/lib64/R/library to libPath be a bit odd? Or simply redundant, since it is the default?

shivaram: So its '../' relative to the directory in which the script is present (we get the script's directory at the top). So for example if sparkR-submit is at `/usr/lib64/R/library/SparkR` then this will add `/usr/lib64/R/library/SparkR/..` to libPaths.

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRRunner.scala
shivaram: Good call - I added this now

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRRunner.scala
shivaram: I think the `try`, `finally` pattern is pretty readable, so I've updated to use this now.

shivaram:
Thanks again @felixcheung -- I've also merged this up with current master branch

pkg/inst/sparkR-submit
felixcheung: Got it, that's what I mean it seems redundant for `/usr/lib64/R/library/SparkR/..`, but it is harmless.

pkg/R/sparkR.R
```diff
@@ -116,16 +116,33 @@ sparkR.init <- function(
...
+  }
 
   .sparkREnv$sparkRBackendPort <- sparkRBackendPort
   cat("Waiting for JVM to come up...\n")
   tries <- 0
-  while(tries < sparkRRetryCount) {
-    if(!connExists(.sparkREnv)) {
+  while (tries < sparkRRetryCount) {
+    if (!connExists(.sparkREnv)) {
       Sys.sleep(2 ^ tries)
```
felixcheung: With the merge from master, we are going to sleep for at least 1 second when we are simply reconnect to backend on the existing port (sparkRExistingPort path on line 122).

It is likely fine as we need to try/retry the connection even in that case. But the sleep might not be absolutely necessary, since it is not launching a new process. Thought?

felixcheung:
LGTM. Thanks!

pkg/R/sparkR.R
shivaram: Its a good point that this adds a 1 sec wait & check even for existing processes -- but I think the overhead should be negligible given this is a one-time thing.  We can come back to fix it later if required.

shivaram:
Alright, I'm merging this change now -- Lets continue testing `sparkR-submit` and see if any other issues come up.

PR #175, opened 2015-02-19
Minor documentation cleanup

falaki:
* We were referencing rJava in readme file. 
* We use Spark 1.3.0 by default
* Users need to install `testthat` package before running tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ R.
 
 ### Requirements
 SparkR requires Scala 2.10 and Spark version >= 0.9.0. Current build by default uses
-Apache Spark 1.3.0. You can also build SparkR against a
+Apache Spark 1.1.0. You can also build SparkR against a
 different Spark version (>= 0.9.0) by modifying `pkg/src/build.sbt`.
 
 ### Package installation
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -97,7 +91,7 @@ To run one of them, use `./sparkR <filename> <args>`. For example:
 
     ./sparkR examples/pi.R local[2]
 
-You can also run the unit-tests for SparkR by running
+You can also run the unit-tests for SparkR by running. You need to install the [testthat](http://cran.r-project.org/web/packages/testthat/index.html) package first.
```
concretevitamin: "
You can also run the unit tests for SparkR by running (you need to install the [testthat](http://cran.r-project.org/web/packages/testthat/index.html) package first)

    install.packages("testthat") # run this in R
    ./run-tests.sh
"

concretevitamin:
Thanks @falaki for catching the glitches. For default Spark version that we pull in, I think we're still using 1.1.0 (see [here](https://github.com/amplab-extras/SparkR-pkg/blob/master/pkg/src/build.sbt#L32))?

shivaram:
Thanks @falaki - LGTM. Merging this.

PR #176, opened 2015-02-19
[SPARKR-193] Retry backend connection if it doesn't come up

lythesia:
Using loop to check connection state and wait a proper&short time interval during each check.

### DIFF
--- a/pkg/R/sparkR.R
+++ b/pkg/R/sparkR.R
@@ -110,9 +110,19 @@ sparkR.init <- function(
                 mainClass = "edu.berkeley.cs.amplab.sparkr.SparkRBackend",
                 args = as.character(sparkRBackendPort),
                 javaOpts = paste("-Xmx", sparkMem, sep = ""))
-  Sys.sleep(2) # Wait for backend to come up
+
+  cat("Waiting JVM bring up ...\n")
+  while(TRUE) {
+    if(!connExists(.sparkREnv)) {
+      Sys.sleep(1)
+      cat(".")
+      connectBackend("localhost", sparkRBackendPort) # Connect to it
+    } else {
+      cat(" ok.\n")
+      break
+    }
+  }
   .sparkREnv$sparkRBackendPort <- sparkRBackendPort
-  connectBackend("localhost", sparkRBackendPort) # Connect to it
 
   if (nchar(sparkHome) != 0) {
     sparkHome <- normalizePath(sparkHome)
--- a/pkg/R/sparkR.R
+++ b/pkg/R/sparkR.R
@@ -90,7 +90,8 @@ sparkR.init <- function(
   sparkExecutorEnv = list(),
   sparkJars = "",
   sparkRLibDir = "",
-  sparkRBackendPort = 12345) {
+  sparkRBackendPort = 12345,
+  sparkRRetryCount = 6) {
 
   if (exists(".sparkRjsc", envir = .sparkREnv)) {
     cat("Re-using existing Spark Context. Please stop SparkR with sparkR.stop() or restart R to create a new Spark Context\n")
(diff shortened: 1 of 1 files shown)


pkg/R/sparkR.R
```diff
@@ -110,9 +110,19 @@ sparkR.init <- function(
                 mainClass = "edu.berkeley.cs.amplab.sparkr.SparkRBackend",
                 args = as.character(sparkRBackendPort),
                 javaOpts = paste("-Xmx", sparkMem, sep = ""))
-  Sys.sleep(2) # Wait for backend to come up
+
+  cat("Waiting JVM bring up ...\n")
```
shivaram: `"Waiting for JVM to come up...\n"` might be better

pkg/R/sparkR.R
```diff
@@ -110,9 +110,19 @@ sparkR.init <- function(
...
+  while(TRUE) {
+    if(!connExists(.sparkREnv)) {
```
shivaram: I am wondering if we implement this using `tryCatch` we wouldn't need the `connExists` call ?

pkg/R/sparkR.R
```diff
@@ -110,9 +110,19 @@ sparkR.init <- function(
                 mainClass = "edu.berkeley.cs.amplab.sparkr.SparkRBackend",
                 args = as.character(sparkRBackendPort),
                 javaOpts = paste("-Xmx", sparkMem, sep = ""))
-  Sys.sleep(2) # Wait for backend to come up
+
+  cat("Waiting JVM bring up ...\n")
+  while(TRUE) {
```
felixcheung: It should probably fail & give up after a few minutes instead of spinning forever?

pkg/R/sparkR.R
felixcheung: If we are passing .sparkREnv, should it set the sparkRBackendPort like on line 125 first?

lythesia: @shivaram  We can use `tryCatch` instead, though `.sparkRcon` will be set (for `connExists` to check) once the connection is setup sucessfully.
@felixcheung I think we can set it somewhere else, I just keep it unchanged there as before.

pkg/R/sparkR.R
lythesia: That may affect a lot. We can increase the time the thtread sleeps each loop or set a `max_tires` to fail it if reach this limit. Maybe we need another customizable option here?

pkg/R/sparkR.R
lythesia: I'll fix it with other changes together.

pkg/R/sparkR.R
shivaram: Yeah it should be fine to double the wait time, retry 6 times and then exit (so wait times will be 1, 2, 4, 8, 16, 32 etc. giving the backend 1 minute to come up). We could add a new config option, but I'd like to avoid having too many config options if possible

pkg/R/sparkR.R
shivaram: Yeah, so I think the contract we are assuming here is that if `connectBackend` succeeds then it sets a valid `.sparkRCon` in `.sparkREnv`. So I think thats fine to assume that. 

I think @felixcheung 's suggestion of moving line 125 (old line 114) above this block would be good to do even though it doesn't affect correctness here.

pkg/R/sparkR.R
felixcheung: +1 @shivaram

pkg/R/sparkR.R
lythesia: @shivaram  Yes, but `.sparkREnv$sparkRBackendPort` is used in `sparkR.stop`, so I think leave it here is necessary and safe. (Or move it into `connectBackend` ?)

shivaram: Its good to keep it here, but I was just suggesting putting it above this while loop so that, lets say if connectBackend fails and we don't finish the init process. Now if someone calls `sparkR.stop` its good to have the port around to try and stop the backend (if possible).

lythesia:
I simply use `stop` when it failes to connect for now.

shivaram:
Thanks @lythesia  - The updates look good to me. I also tried this out locally and tested it by injecting some delays etc. and the patch works as expected.

@felixcheung Any other comments ? If not I'll merge this.

felixcheung:
LGTM. Thanks for including me.

shivaram:
Great. Merging this

PR #177, opened 2015-02-24
[SPARKR-152] Support functions to change number of RDD partitions (coalesce, repartition)

lythesia:
Add support for `coalesce` and `repartition`, simple tests included.


pkg/R/RDD.R
```diff
@@ -1039,6 +1039,60 @@ setMethod("keyBy",
...
+#' rdd <- parallelize(sc, list(1, 2, 3, 4, 5), 3L)
+#' numPartitions(rdd)               # 3
+#' numPartitions(coalesce(rdd, 1L)) # 1
+#'}
+setGeneric("coalesce", function(rdd, numPartitions, ...) { standardGeneric("coalesce") })
+
+#' @rdname coalesce
+#' @aliases coalesce,RDD
+setMethod("coalesce",
+           signature(rdd = "RDD", numPartitions = "integer"),
```
hlin09: Does it make sense to accept `numPartitions` as "numeric" types, so that user can pass in something like `2` instead of `2L`? We can just cast them to integer inside.

lythesia: Surely we can implement as cast for easy use, but I think it's better to remind user that `numPartitions` should be an integer.

shivaram: Actually I agree with @hlin09 that having an integer argument is not ideal. This is mainly because when users try to use the function with say `2` as the second argument, they get an error which is hard to understand because of the way R's dispatch works (See an example for collectPartition below).

```
unable to find an inherited method for function ‘collectPartition’ for signature ‘"RDD", "numeric"’
```

I think the right thing to do is to add a check to see if the argument is not integer and warn the user. Something like
```
if (as.integer(numPartitions) != numPartitions) {
  warning("numPartitions should be an integer. Coercing to integer")
}
```

pkg/R/RDD.R
```diff
@@ -1039,6 +1039,60 @@ setMethod("keyBy",
...
+#' numPartitions(repartition(rdd, 2L))  # 2
+#'}
+setGeneric("repartition", function(rdd, numPartitions) { standardGeneric("repartition") })
+
+#' @rdname repartition
+#' @aliases repartition,RDD
+setMethod("repartition",
+          signature(rdd = "RDD", numPartitions = "integer"),
+          function(rdd, numPartitions) {
+            jrdd <- callJMethod(getJRDD(rdd), "repartition", numPartitions)
```
shivaram: I'm not sure this works as required. This is mainly because in SparkR if the data is encoded as `byte` we only have 1 record in each partition. (The entire partition is encoded as a single byte array). So when the JVM tries to hash keys to partitions we will end up with entire partitions moving together. 

I think the way to get around this is to create a PairRDD just like its created in Scala [1] and then call `partitionBy` and pass in the number of partitions.
 
[1] https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/rdd/RDD.scala#L358

pkg/R/RDD.R
```diff
@@ -1039,6 +1039,60 @@ setMethod("keyBy",
...
+#' numPartitions(coalesce(rdd, 1L)) # 1
+#'}
+setGeneric("coalesce", function(rdd, numPartitions, ...) { standardGeneric("coalesce") })
+
+#' @rdname coalesce
+#' @aliases coalesce,RDD
+setMethod("coalesce",
+           signature(rdd = "RDD", numPartitions = "integer"),
+           function(rdd, numPartitions, shuffle = FALSE) {
+             jrdd <- callJMethod(getJRDD(rdd), "coalesce", numPartitions, shuffle)
```
shivaram: Similar to the comment in repartition if the number of partitions is increasing we will need to do the shuffle as described before. If the number of partitions is reducing I think we can just call the JVM `coalesce` method.

shivaram:
@lythesia Thanks for sending out the change. I left some comments inline. In addition I just wanted to list out the command I used for testing.

```
a <- parallelize(sc, 1:20, 2L)
jrdd <- SparkR:::callJMethod(SparkR:::getJRDD(a), "repartition", 4L)
b <- SparkR:::RDD(jrdd)
collectPartition(b, 0L)  # Should contain 4 numbers but is empty on my machine
collectPartition(b, 1L)
collectPartition(b, 2L)
```

pkg/R/RDD.R
lythesia: Ok, I see that and I'll fix it.
Two more questions: when generating random partition keys(total # = `numPartitions`), some keys will be skipped occasionally, which leads to empty partition (though it is quite impossible), does it matters? Should we ensure each partition has a different random seed just like [sampleRDD](https://github.com/amplab-extras/SparkR-pkg/blob/master/pkg/R/RDD.R#L911)?

shivaram: We can just use the partitionId as the seed inside `mapPartitionsWithIndex`. I think that should ensure a different random seed in each partition.

pkg/R/RDD.R
```diff
@@ -1039,6 +1039,77 @@ setMethod("keyBy",
...
+#' @seealso coalesce
+#' @export
+#' @examples
+#'\dontrun{
+#' sc <- sparkR.init()
+#' rdd <- parallelize(sc, list(1, 2, 3, 4, 5, 6, 7), 4L)
+#' numPartitions(rdd)                   # 4
+#' numPartitions(repartition(rdd, 2L))  # 2
+#'}
+setGeneric("repartition", function(rdd, numPartitions) { standardGeneric("repartition") })
```
davies: We had agreed that use `x` in signature for all generic method.

lythesia:
@shivaram Sorry for misunderstand of random partition key generation, now I think repartition distributes elements evenly into new partitions.
Since R has no native apply-with-index, I used a dirty `seq_along`, better suggestions?

pkg/inst/tests/test_rdd.R
```diff
@@ -267,6 +267,20 @@ test_that("keyBy on RDDs", {
   expect_equal(actual, lapply(nums, function(x) { list(func(x), x) }))
 })
 
+test_that("repartition/coalesce on RDDs", {
+  rdd <- parallelize(sc, 1:10, 3L)
+  expect_equal(numPartitions(rdd), 3L)
```
shivaram: Could you also check how many elements are there in each partition using `collectPartition` ? Its okay to have it be a range (i.e. less than 5 and greater than 2 etc.) if its not deterministic.

pkg/R/RDD.R
```diff
@@ -1039,6 +1039,78 @@ setMethod("keyBy",
...
+                 set.seed(s)  # split as seed
+                 start <- as.integer(runif(1, 0, numPartitions))
+                 lapply(seq_along(part),
+                        function(i) {
+                          pos <- (start + i) %% numPartitions
+                          list(pos, part[[i]])
+                        })
+               }
+               shuffled <- lapplyPartitionsWithIndex(x, func)
+               reparted <- partitionBy(shuffled, numPartitions)
```
shivaram: `repartitioned` might be a better name than `reparted`

lythesia:
Add range checks for new partition.

pkg/R/RDD.R
```diff
@@ -1039,6 +1039,78 @@ setMethod("keyBy",
...
+#' numPartitions(coalesce(rdd, 1L)) # 1
+#'}
+setGeneric("coalesce", function(x, numPartitions, ...) { standardGeneric("coalesce") })
+
+#' @rdname coalesce
+#' @aliases coalesce,RDD
+setMethod("coalesce",
+           signature(x = "RDD", numPartitions = "numeric"),
+           function(x, numPartitions, shuffle = FALSE) {
+             if(as.integer(numPartitions) != numPartitions) {
```
shivaram: space after `if`

pkg/R/RDD.R
```diff
@@ -1039,6 +1039,78 @@ setMethod("keyBy",
...
+              warning("Number of partitions should be an integer. Coercing it to integer.")
+             }
+             numPartitions <- as.integer(numPartitions)
+             if (shuffle || numPartitions > SparkR::numPartitions(x)) {
+                func <- function(s, part) {
+                 set.seed(s)  # split as seed
+                 start <- as.integer(runif(1, 0, numPartitions))
```
shivaram: instead of `runif` we can use `sample(numPartitions, 1) - 1` here -- (sample chooses from 1 to numPartitions, so we need a -1)

shivaram:
Thanks @lythesia  LGTM

pkg/R/RDD.R
```diff
@@ -1039,6 +1039,78 @@ setMethod("keyBy",
...
+#' @aliases coalesce,RDD
+setMethod("coalesce",
+           signature(x = "RDD", numPartitions = "numeric"),
+           function(x, numPartitions, shuffle = FALSE) {
+             if (as.integer(numPartitions) != numPartitions) {
+              warning("Number of partitions should be an integer. Coercing it to integer.")
+             }
+             numPartitions <- as.integer(numPartitions)
+             if (shuffle || numPartitions > SparkR::numPartitions(x)) {
+                func <- function(s, part) {
```
davies: indent with one less space

davies:
LGTM

shivaram:
Great. Thanks for the fixes. Merging this

PR #178, opened 2015-02-25
[SPARKR-204] use random port in backend

davies:
Backend will bind to random port, and send back to R via another socket.

### DIFF
--- a/pkg/R/sparkR.R
+++ b/pkg/R/sparkR.R
@@ -88,9 +88,7 @@ sparkR.init <- function(
   sparkEnvir = list(),
   sparkExecutorEnv = list(),
   sparkJars = "",
-  sparkRLibDir = "",
-  sparkRBackendPort = as.integer(Sys.getenv("SPARKR_BACKEND_PORT", "12345")),
-  sparkRRetryCount = 6) {
+  sparkRLibDir = "") {
 
   if (exists(".sparkRjsc", envir = .sparkREnv)) {
     cat("Re-using existing Spark Context. Please stop SparkR with sparkR.stop() or restart R to create a new Spark Context\n")
@@ -121,45 +119,38 @@ sparkR.init <- function(
   if (sparkRExistingPort != "") {
     sparkRBackendPort <- sparkRExistingPort
   } else {
+    # TODO: test the random port and retry, or use httpuv:::startServer
+    callbackPort = sample(40000, 1) + 10000
     if (Sys.getenv("SPARKR_USE_SPARK_SUBMIT", "") == "") {
       launchBackend(classPath = cp,
                     mainClass = "edu.berkeley.cs.amplab.sparkr.SparkRBackend",
-                    args = as.character(sparkRBackendPort),
+                    args = as.character(callbackPort),
                     javaOpts = paste("-Xmx", sparkMem, sep = ""))
     } else {
       # TODO: We should deprecate sparkJars and ask users to add it to the
       # command line (using --jars) which is picked up by SparkSubmit
       launchBackendSparkSubmit(
           mainClass = "edu.berkeley.cs.amplab.sparkr.SparkRBackend",
-          args = as.character(sparkRBackendPort),
+          args = as.character(callbackPort),
           appJar = .sparkREnv$assemblyJarPath,
           sparkHome = sparkHome,
           sparkSubmitOpts = Sys.getenv("SPARKR_SUBMIT_ARGS", ""))
     }
+    sock <- socketConnection(port = callbackPort, server = TRUE, open = 'rb',
+        blocking = TRUE, timeout = 10)
+    sparkRBackendPort <- readInt(sock)
+    if (length(sparkRBackendPort) == 0) {
+      stop("JVM failed to launch")
+    }
+    close(sock)
   }
 
   .sparkREnv$sparkRBackendPort <- sparkRBackendPort
-  cat("Waiting for JVM to come up...\n")
-  tries <- 0
-  while (tries < sparkRRetryCount) {
-    if (!connExists(.sparkREnv)) {
-      Sys.sleep(2 ^ tries)
-      tryCatch({
-        connectBackend("localhost", .sparkREnv$sparkRBackendPort)
-      }, error = function(err) {
-        cat("Error in Connection, retrying...\n")
-      }, warning = function(war) {
-        cat("No Connection Found, retrying...\n")
-      })
-      tries <- tries + 1
-    } else {
-      cat("Connection ok.\n")
-      break
-    }
-  }
-  if (tries == sparkRRetryCount) {
-    stop(sprintf("Failed to connect JVM after %d tries.\n", sparkRRetryCount))
-  }
+  tryCatch({
+    connectBackend("localhost", sparkRBackendPort)
+  }, error = function(err) {
+    stop("Failed to connect JVM\n")
+  })
 
   if (nchar(sparkHome) != 0) {
     sparkHome <- normalizePath(sparkHome)
--- a/pkg/R/sparkR.R
+++ b/pkg/R/sparkR.R
@@ -135,11 +135,14 @@ sparkR.init <- function(
           sparkHome = sparkHome,
           sparkSubmitOpts = Sys.getenv("SPARKR_SUBMIT_ARGS", ""))
     }
-    for (i in 1:100) {
+    # wait atmost 100 seconds for JVM to launch 
+    wait <- 0.1
+    for (i in 1:25) {
+      Sys.sleep(wait)
       if (file.exists(path)) {
         break
       }
-      Sys.sleep(0.1)
+      wait <- wait * 1.25
     }
     if (!file.exists(path)) {
       stop("JVM is not ready after 10 seconds")
--- a/pkg/R/sparkR.R
+++ b/pkg/R/sparkR.R
@@ -119,30 +119,38 @@ sparkR.init <- function(
   if (sparkRExistingPort != "") {
     sparkRBackendPort <- sparkRExistingPort
   } else {
-    # TODO: test the random port and retry, or use httpuv:::startServer
-    callbackPort = sample(40000, 1) + 10000
+    path <- tempfile(pattern = "backend_port")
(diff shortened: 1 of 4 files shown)


davies:
The failure seems not related to this PR, how to re-test it?

shivaram:
I can retrigger the travis job -- BTW do the tests pass on your machine ? You can run it with `./run-tests.sh`

davies:
Yes, they all passed.

pkg/R/sparkR.R
```diff
@@ -121,45 +119,38 @@ sparkR.init <- function(
...
       # command line (using --jars) which is picked up by SparkSubmit
       launchBackendSparkSubmit(
           mainClass = "edu.berkeley.cs.amplab.sparkr.SparkRBackend",
-          args = as.character(sparkRBackendPort),
+          args = as.character(callbackPort),
           appJar = .sparkREnv$assemblyJarPath,
           sparkHome = sparkHome,
           sparkSubmitOpts = Sys.getenv("SPARKR_SUBMIT_ARGS", ""))
     }
+    sock <- socketConnection(port = callbackPort, server = TRUE, open = 'rb',
```
shivaram: Can we bring up this server socket before we launch the backend ? I just want to avoid the race condition where the backend tries to open a socket and the server socket was not created for some reason.

pkg/R/sparkR.R
```diff
@@ -121,45 +119,38 @@ sparkR.init <- function(
   if (sparkRExistingPort != "") {
     sparkRBackendPort <- sparkRExistingPort
   } else {
+    # TODO: test the random port and retry, or use httpuv:::startServer
+    callbackPort = sample(40000, 1) + 10000
```
shivaram: Can we have a simple loop trying a few increments ? (Something similar to the connection retry we used to have). It would also be good to `stop` with an error message if the `socketConnection` doesn't come up successfully after the retries.

pkg/R/sparkR.R
davies: This problem is that this line will block until a client connection come up, so we can not move it before launch JVM (it's weird that R does not have a fully functional socket library).

There is a race condition here, but given the fact that JVM usually take 1-2 seconds to startup, the risk is very low. We can see it actual cause problem or not.

pkg/R/sparkR.R
davies: We pick the port first, and pass it into JVM, then we can not try and change the port anymore.

The final solution maybe use httpuv to launch a http server in the background with random port, then pass the real port into JVM, JVM send back the port of backend via HTTP request.

pkg/R/sparkR.R
shivaram: Hmm I see -- Yeah, so R is single-threaded completely and thats why I guess it blocks for a connection to come up.
BTW, can we use the JVM process's `stdout`or a known filename instead ? I know that recently in PySpark you went from `stdout` to a socket based solution -- was there any reasons for that ?

davies: stdout and stderr is not safe to communicate, any logging will introduce troubles.

--  
Davies Liu
Sent with Sparrow (http://www.sparrowmailapp.com/?sig)

shivaram: Hmm what about this solution -- We pass in a temp file name as the first argument to SparkRRunner which writes out the port number to that file. We can use `tempfile`[1] to get a unique file name and the one complexity is that we might need to retry the file read.

I'm just trying to see if we can avoid the risk of picking a port that is already used in `socketConnection` or of the JVM race condition etc.


[1] https://stat.ethz.ch/R-manual/R-patched/library/base/html/tempfile.html

davies: Good idea, will do.

pkg/R/sparkR.R
```diff
@@ -121,45 +119,46 @@ sparkR.init <- function(
...
+          args = path,
           appJar = .sparkREnv$assemblyJarPath,
           sparkHome = sparkHome,
           sparkSubmitOpts = Sys.getenv("SPARKR_SUBMIT_ARGS", ""))
     }
+    for (i in 1:100) {
+      if (file.exists(path)) {
+        break
+      }
+      Sys.sleep(0.1)
```
shivaram: We can make this exponential back-off multiplying each time by 2 (till say 10 iterations ?). That'll also give the JVM around 1 minute to come up which might happen if the process launch + port binding is slow.

pkg/R/sparkR.R
```diff
@@ -121,45 +119,46 @@ sparkR.init <- function(
...
+    }
+    if (!file.exists(path)) {
+      stop("JVM is not ready after 10 seconds")
+    }
+    f <- file(path, open='rb')
+    sparkRBackendPort <- readInt(f)
+    close(f)
+    file.remove(path)
```
shivaram: You can just use `unlink(f)` instead of close + remove

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackend.scala
```diff
@@ -80,18 +78,28 @@ class SparkRBackend {
...
     }
     val sparkRBackend = new SparkRBackend()
     try {
-      sparkRBackend.init(args(0).toInt)
+      // bind to random port
+      val boundPort = sparkRBackend.init()
+      // tell the R process via temporary file
+      val path = args(0)
+      val f = new File(path + ".tmp")
+      val output = new FileOutputStream(f)
```
shivaram: Can combine this into one line `val dos = new DataOutputStream(FileOutputStream(f))` ?

shivaram:
Thanks for the update. The new method looks good to me and I had some minor comments inline. However this PR keeps failing the Travis build and I don't know why. I pulled and tested locally too and it worked fine. I've emailed the Travis support to see if they can help us debug.

pkg/R/sparkRClient.R
```diff
@@ -2,7 +2,7 @@
 
 # Creates a SparkR client connection object
 # if one doesn't already exist
-connectBackend <- function(hostname, port, timeout = 6000) {
+connectBackend <- function(hostname, port, timeout = 6) {
```
shivaram: @davies So after some debugging in a Ubuntu VM, I found the reason for the travis test failures. This timeout here is not just timeout to make the connection but is the timeout for any messages on the socket. So if we say call `collect` using `callJMethod` then we wait on the socket till there is a reply. Now for long running tasks this may take longer than 6 seconds. I set it to 6000 previously as an upper bound on the time a call will take to return. Its not very ideal, but without having a polling based system or some callback like systems I am not sure its easy to fix. 

Anyways lets not modify this timeout in this PR, we can revisit this later.

pkg/R/sparkR.R
davies: unlink() is similar to file.remove(f), you still need to close it first (tested on Mac).

pkg/R/sparkR.R
davies: The JVM usually take more than 1 seconds to launch. If we increase the period time by 2, we will wail about 1 second to find out that the file is ready.

We can expect that the JVM will take one to several seconds, then the current approach is better (wake up as soon as possible).

pkg/R/sparkRClient.R
davies: Good point, I will change it back.

Should we use blocking mode? It's possible that a job may take a few hours. In PySpark, it's in blocking mode.

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackend.scala
davies: will do.

pkg/R/sparkR.R
shivaram: Yeah but I am more worried about the cases where it takes longer than 10 seconds. Sometimes in the travis build and also in low memory containers I've seen longer delays. So I'd like to wait up to 1 minute in total and in that case checking every 0.1 seconds seems very wasteful. 

We could use a lower factor than 2, say 1.5 -- In that case if we use 13 iterations of the loop the times at which we check would be
```
0.150  0.375  0.713  1.219  1.978  3.117  4.826  7.389 11.233 17.000 25.649 38.624 58.086
```

So we'd check at 0.7, 1, 2, 3 seconds (covers common cases) but also handle cases up to 1 minute long.

felixcheung:
+1 nice. LGTM.

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackend.scala
```diff
@@ -80,18 +78,26 @@ class SparkRBackend {
 object SparkRBackend {
   def main(args: Array[String]) {
     if (args.length < 1) {
-      System.err.println("Usage: SparkRBackend <port>")
+      System.err.println("Usage: SparkRBackend <path>")
```
concretevitamin: Could we give it a more descriptive name? e.g. "tempFilePath"

concretevitamin:
LGTM pending a minor style comment - BTW Travis is happy now.

pkg/R/sparkR.R
concretevitamin: One concern is that 1s of developer time is much more valuable than 1min of machine spinning time. Having SparkR shell start up ASAP is important for user experience (we don't want it to feel laggy).

shivaram: Yeah but there is some trade-off if we want to handle wait times up to 60 seconds (due to a slow machine) and keeping it responsive. Checking every 0.1 second would result in 600 system calls (`file.exists` needs to call `stat` I'd think) so we need to balance this out. 

Actually if 1 second wait granularity is fine, then we could just check every second 60 times. Do you think that is better than exponential back-off ?

concretevitamin: I guess 0.5s is better, but also okay with every 1s - @davies?

shivaram:
Thanks @davies for the update. The latest changes LGTM. 

@concretevitamin  Any further comments ?

concretevitamin:
LGTM - merging this

PR #179, opened 2015-02-25
[SPARKR-208] Changed SQLContext.R#parquetFile signature to match the java version

evertlammerts:
This is a simple fix to make parquetFile work in R. See https://sparkr.atlassian.net/browse/SPARKR-208 for details.

### DIFF
--- a/pkg/R/SQLContext.R
+++ b/pkg/R/SQLContext.R
@@ -62,17 +62,15 @@ jsonRDD <- function(sqlCtx, rdd, schema = NULL, samplingRatio = 1.0) {
 #' Loads a Parquet file, returning the result as a DataFrame.
 #'
 #' @param sqlCtx SQLContext to use
-#' @param path Path of file to read. A vector of multiple paths is allowed.
+#' @param ... Path(s) of parquet file(s) to read.
 #' @return DataFrame
 #' @export
 
 # TODO: Implement saveasParquetFile and write examples for both
-parquetFile <- function(sqlCtx, path) {
+parquetFile <- function(sqlCtx, ...) {
   # Allow the user to have a more flexible definiton of the text file path
-  path <- normalizePath(path)
-  # Convert a string vector of paths to a string containing comma separated paths
-  path <- paste(path, collapse = ",")
-  sdf <- callJMethod(sqlCtx, "parquetFile", path)
+  paths <- lapply(list(...), normalizePath)
+  sdf <- callJMethod(sqlCtx, "parquetFile", paths)
   dataFrame(sdf)
 }
 
--- a/pkg/inst/tests/test_sparkSQL.R
+++ b/pkg/inst/tests/test_sparkSQL.R
@@ -527,15 +527,23 @@ test_that("withColumn() and withColumnRenamed()", {
   expect_true(columns(newDF2)[1] == "newerAge")
 })
 
-# TODO: Enable and test once the parquetFile PR has been merged
-# test_that("saveAsParquetFile() on DataFrame and works with parquetFile", {
-#   df <- jsonFile(sqlCtx, jsonPath)
-#   parquetPath <- tempfile(pattern="spark-test", fileext=".tmp")
-#   saveAsParquetFile(df, parquetPath)
-#   parquetDF <- parquetFile(sqlCtx, parquetPath)
-#   expect_true(inherits(parquetDF, "DataFrame"))
-#   expect_equal(collect(df), collect(parquetDF))
-#   unlink(parquetPath)
-# })
+test_that("saveDF() on DataFrame and works with parquetFile", {
+  df <- jsonFile(sqlCtx, jsonPath)
+  saveDF(df, parquetPath, "parquet", mode="overwrite")
+  parquetDF <- parquetFile(sqlCtx, parquetPath)
+  expect_true(inherits(parquetDF, "DataFrame"))
+  expect_equal(collect(df), collect(parquetDF))
+})
 
+test_that("parquetFile works with multiple input paths", {
+  df <- jsonFile(sqlCtx, jsonPath)
+  saveDF(df, parquetPath, "parquet", mode="overwrite")
+  parquetPath2 <- tempfile(pattern = "parquetPath2", fileext = ".parquet")
+  saveDF(df, parquetPath2, "parquet", mode="overwrite")
+  parquetDF <- parquetFile(sqlCtx, parquetPath, parquetPath2)
+  expect_true(inherits(parquetDF, "DataFrame"))
+  expect_true(count(parquetDF) == count(df)*2)
+})
+
+unlink(parquetPath)
 unlink(jsonPath)


pkg/R/SQLContext.R
```diff
@@ -36,12 +36,14 @@ jsonFile <- function(sqlCtx, path) {
...
-  # Convert a string vector of paths to a string containing comma separated paths
-  path <- paste(path, collapse=",")
-  sdf <- callJMethod(sqlCtx, "parquetFile", path)
+parquetFile <- function(sqlCtx, ...) {
+  # Iterate through paths
+  paths <- list(...)
+  for (i in seq_along(paths)) {
+    # Allow the user to have a more flexible definiton of the text file path
+    paths[[i]] <- normalizePath(paths[[i]])
+  }
```
cafreeman: This looks good. Interesting that our standard approach of passing a comma-separated string of paths doesn't work in this case. @shivaram is this a new change that might affect our other read methods? I think almost all of them use the `paste(path, collapse=",")` trick to handle multiple files.

cafreeman:
@evertlammerts Thanks for taking a look at this! Looking through the current PySpark method, your changes are definitely in line with the existing approaches.

I left a comment in-line and wouild be curious to hear your take as well.

pkg/R/SQLContext.R
```diff
@@ -36,12 +36,14 @@ jsonFile <- function(sqlCtx, path) {
...
 # TODO: Implement saveasParquetFile and write examples for both
-parquetFile <- function(sqlCtx, path) {
-  # Allow the user to have a more flexible definiton of the text file path
-  path <- normalizePath(path)
-  # Convert a string vector of paths to a string containing comma separated paths
-  path <- paste(path, collapse=",")
-  sdf <- callJMethod(sqlCtx, "parquetFile", path)
+parquetFile <- function(sqlCtx, ...) {
+  # Iterate through paths
+  paths <- list(...)
```
concretevitamin: I think this would be simpler: 
```
paths <- lapply(list(...), normalizePath)
```

concretevitamin:
Thanks @evertlammerts - Looking at the ticket, I think generally I'm +1 on having SparkR's interface match with existing Scala/Python ones,

pkg/R/SQLContext.R
shivaram: I think the reason it doesn't work is that the Java method signature looks different for `parquetFile` when compared to something like `textFile`. One thing we could do is add unit tests with multiple files for all read methods so the code paths get tested better.

evertlammerts:
@concretevitamin definitely better with lapply!

pkg/R/SQLContext.R
```diff
@@ -36,12 +36,10 @@ jsonFile <- function(sqlCtx, path) {
 #' @export
```
shivaram: @evertlammerts Could you also update the doc above this to indicate `...` is for paths to parquet files ?

evertlammerts: Right, done

evertlammerts:
@davies in edu.berkeley.cs.amplab.sparkr.SerDe it looks like lists are mapped onto Array[T]. I've only tested with a single argument (i.e. parquetFile(cqlCtx, path)) and that seemed to work. Can't test multiple paths right now. Are you suggesting SerDe should be changed?

davies:
@evertlammerts I think we can't, because some Java API accept Array[T], we could manually create Seq for varargs.

davies:
see https://github.com/amplab-extras/SparkR-pkg/pull/180/files#diff-d3e9e85faf4287e4eabba38c49232a9dR19

shivaram:
Thanks @evertlammerts for the update. I think @davies point is that if some methods only take in Seq then we can use a Array to Seq utility method to do the conversion. In this case if varargs works it should be fine.

Also we should add test cases for `parquetFile` (and one with multiple file names as well). @evertlammerts do you have a small parquet file we can add to the tests directory and use for testing ? If not I'll file a new JIRA to add this when we have `saveAsParquetFile` implemented.

evertlammerts:
@shivaram sure, how about this? Created with SchemaRDD#saveAsParquetFile. 5 cols (ints), 20 rows. Does SparkR include tests? I can't seem to find them so I'm not sure where to add this to other than pkg/test... 

@davies Agreed, it would be good to handle varargs in SparkRBackendHandler#matchMethod. I'll see whether I can come up something.

evertlammerts:
What do you think of it like this, @davies and @shivaram? I'm not sure how to test other than through parquetFile, and that seems to work with multiple paths if I try. Maybe add some tests for this?

I've used Array#toSeq because your changes are not in yet. SQLUtils#toSeq would work better.

shivaram:
Thanks @evertlammerts for the update. The parquet file created sounds good to me. The SparkR test files are in `pkg/inst/tests` and specifically the SparkSQL tests cases are in [1]. You can add a new test case here where you say open a parquet file and run count on it.

Also I think the parquet file should be in pkg/inst/tests as then the test cases should be able to pick it up without any path manipulation. Let me know if you have any problems with this.

Regarding the Seq vs. varargs, I am not in favor of adding more logic to the SparkRBackendHandler if possible. Since @davies 's arrayToSeq function was pretty small, do you mind including that here ? (We can handle conflicts during merge).

[1] https://github.com/amplab-extras/SparkR-pkg/blob/sparkr-sql/pkg/inst/tests/test_sparkSQL.R

evertlammerts:
FYI: I had been testing against Spark master and only see now that somewhere after 1.2.1 the signature of SQLContext#parquetFile was changed to (path: String, paths: String*), and ten days ago it was changed back to (paths: String*): https://github.com/apache/spark/commit/0e180bfc3c7f18780d4fc4f42681609832418e43. Testing against the Spark version in apache's maven repo fails because of this.

shivaram:
Thanks @evertlammerts - I didn't know that. Lets use whatever is in `branch-1.3` as the reference for now. I think a new RC will be cut soon, so we should be fine depending on that.

@davies can probably tell us if the API has been finalized for parquetFile

cafreeman:
@shivaram building from `branch-1.3

shivaram:
@cafreeman Thanks for the heads up. I will try and see why 1.3.1-SNAPSHOT doesn't work.  My comment was just that in terms of interface to `parquetFile` we can see what the 1.3 release is going to look like in branch-1.3

shivaram:
@evertlammerts FYI we just recently moved to using a more recent snapshot of Spark 1.3 in our build https://github.com/amplab-extras/SparkR-pkg/pull/195 -- I think it should be possible to update this to the stable version of `parquetFile` now

evertlammerts:
Alright, thanks for the heads up. I think I'll have time to give that a try
tomorrow.

shivaram:
Thanks for the update @evertlammerts  -- I hate to chain PRs but it might be much easier to add a test case for this once https://github.com/amplab-extras/SparkR-pkg/pull/199 gets checked in. That adds support for `saving` DataFrames to parquet files so we can test this PR by just creating a parquet file from a JSON RDD. Let me know if that sounds good.

evertlammerts:
Sounds good to me, @shivaram. This change seems to work with `1.3.0-rc2`, but a bunch of testcases in `test_sparkSQL.R` are failing for me. I'll wait until #199 is merged.

cafreeman:
I believe the test cases are failing because the `Column` class is still using the old version. We need to merge in the current `sparkr-sql` branch to get all the updates (like having `Column` use `functions` instead of `Dsl`) and then this should successfully compile against the RC2 build.

shivaram:
@evertlammerts Did you get a chance to look at this ? #199 has also been merged.

evertlammerts:
This should pass now. Do you also want saveAsParquetFile tests? I'm using saveDF right now because it allows overwriting existing files, which seems cleaner.

shivaram:
Thanks @evertlammerts. Using saveDF is fine for this PR. We can add saveAsParquetFile tests separately if required.

LGTM

PR #180, opened 2015-02-25
[SPARKR-191] groupBy and agg() API for DataFrame

davies:
TODO: docs

for sum/min/max/avg/mean, there are common functions in R, I'd like to NOT override them for GroupedData.

We could alway do with 
```
agg(gd, name='min', age='max')
```


pkg/R/DataFrame.R
```diff
@@ -276,3 +298,37 @@ setMethod("foreachPartition",
...
+
+############################## GroupedData ########################################
+
+setClass("GroupedData",
+         slots = list(env = "environment",
+                      sgd = "jobj"))
+
+setMethod("initialize", "GroupedData", function(.Object, sgd) {
+  .Object@env <- new.env()
+  .Object@sgd <- sgd
```
cafreeman: Is there a `cache` or `persist` method for GroupedData? If so, we probably want to add an `isCached` flag in `.Object@env`

cafreeman:
@davies This is awesome! Happy to see new pieces getting added to the SparkSQL side of things :)

I have a general request: can you change the method signatures to use `x` instead of `df`? We're currently standardizing all of the RDD/DataFrame method to use `x` so that the signatures don't start to look weird when we have multiple dispatch on different methods.

So, for instance, `groupBy` would become:

```R
setGeneric("groupBy", function(x, ...) { standardGeneric("groupBy") })

setMethod("groupBy",
          signature(x = "DataFrame"),
          function(x, col, ...) {
            jseq <- callJStatic("edu.berkeley.cs.amplab.sparkr.SQLUtils", "toSeq", list(...))
            sgd <- callJMethod(x@sdf, "groupBy", col, jseq)
            groupedData(sgd)
          })
```

pkg/R/DataFrame.R
davies: No, `GroupedData` can not be materialized.

davies:
@cafreeman @shivaram  I'd like to leave the docs out of this PR, then I can help from others to fix the docs, or fill them in batch before merging into master. Does it work for you?

cafreeman:
I think it's better to have the roxygen2 documentation in the code before merging. I'm indifferent as to whether we actually generate the .Rd files at this point, but I think at least having the comments in place keeps us from letting the lack of documentation get out of hand.

davies:
Currently, the complete version of docstring looks too verbose, it make a 5 line method becoming more than 20 lines.

The whole DataFrame API are still WIP, so I think we could focus on the code and tests first, then focus on docs later (it's easy to add docs in batch).

cafreeman:
I defer to @shivaram and @concretevitamin on this one. Whatever works for them is fine by me.

pkg/R/DataFrame.R
```diff
@@ -276,3 +299,58 @@ setMethod("foreachPartition",
...
+              jseq <- callJStatic("edu.berkeley.cs.amplab.sparkr.SQLUtils", "toSeq", list(...))
+              sdf <- callJMethod(x@sgd, name, jseq)
+              dataFrame(sdf)
+            })
+}
+
+for (name in metheds) {
+  createMethod(name)
+}
+
```
concretevitamin: Remove extra blank lines?

pkg/NAMESPACE
```diff
@@ -84,8 +84,14 @@ exportClasses("DataFrame")
 exportMethods("printSchema",
               "registerTempTable",
               "toRDD",
+              "groupBy",
+              "agg",
               "limit")
 
+exportClasses("GroupedData")
+exportMethods("count", "agg", "avg")
+
```
concretevitamin: extra blank line

concretevitamin:
@davies I think leaving the docs out for now is fair; let's file a JIRA for it.

davies:
Created one : https://sparkr.atlassian.net/browse/SPARKR-213

pkg/NAMESPACE
davies: I'd like to keep them close, they are in same group: `GroupedData`

davies:
Because the spark-sql-1.3.0 in maven is outdated, the sum/max/min API are written against latest spark-sql, so they can not passwd the tests, we can hold this until new spark-sql is published into maven.

shivaram:
@davies Its okay to add documentation later but I want to think a little bit more about the interface. I see that the current interface is using named arguments similar in some sense to Python API. I am wondering if we can do something a little more R-friendly here.

On that note, it'll be good to add a couple of example usages on top of the methods. Need not be a full end-to-end example but just to show syntax.

Regarding the API, one dataframe API that is pretty popular in R is the `dplyr` API described at http://cran.rstudio.com/web/packages/dplyr/vignettes/introduction.html -- In `dplyr` aggregations are done
as `summarise(destinations, flights = count(tailnum))` where `destinations` is a data frame and `flights` is the column created with the counts. I think this API is probably easier to read than the `tailnum = count` version we have now (btw how do we give a new column name ?) but it might be trickier to implement this.

@dputler @cafreeman @concretevitamin -- Any thoughts on the interface for aggregation ?

cafreeman:
I absolutely agree on the `dplyr` suggestion. My goal is to model the API on any of the similar functions found in `dplyr` and, to a lesser degree, `tidyr`. A good one-off snapshot of what this looks like can be found here:

http://www.rstudio.com/wp-content/uploads/2015/02/data-wrangling-cheatsheet.pdf

davies:
@shivaram Once we have Column, we could have the API as you suggested:

```
agg(gd, count(column_name))
agg(gd, new_column = count(column_name))
```

shivaram:
@davies That's great - FWW I just remembered that one can do this even without column support as R allows you to capture an argument and convert it to string without evaluating it.  For example
```
agg <- function(df, aggF) { 
  cat("aggregating ", deparse(substitute(df)), " using ", deparse(substitute(aggF)), "\n") 
}
agg(peopleDF, mean(age))
```

The `deparse(substitute())` functions read in the argument `aggF` and converts it to a character string without evaluating it.

BTW do we want to hold off on this and merge it after the column PR ? or do you want to merge this and then further APIs with the Column PR ?

davies:
I'd like to hold off on this, until we can build with a new version of spark sql 1.3

cafreeman:
We could revert back to using `1.3.0-SNAPSHOT` in `build.sbt`. We'd just all have to clone the most up-to-date `master` branch of Spark. I tried it with `branch-1.3` but it builds to `1.3.1` and the SparkRbackend seems to freak out for some reason.

dputler:
Chiming in late, yes I think that following the dplyr interface makes sense, given the traction of that project. Having said this, we probably want to make sure we maintain the old school way of doing things, which for aggregates, generally means using lapply.

 

Dan

pkg/R/DataFrame.R
```diff
@@ -276,3 +299,54 @@ setMethod("foreachPartition",
...
+          signature(x = "GroupedData"),
+          function(x, ...) {
+            cols <- varargsToEnv(...)
+            sdf <- callJMethod(x@sgd, "agg", cols)
+            dataFrame(sdf)
+          })
+
+#' sum/mean/avg/min/max
+
+metheds <- c("sum", "mean", "avg", "min", "max")
```
cafreeman: spelling nit: "methods"

cafreeman:
Anyone know where we are on this one?

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
```diff
@@ -160,6 +166,7 @@ class SparkRBackendHandler(server: SparkRBackend) extends SimpleChannelInboundHa
         }
       }
       if (!parameterWrapperType.isInstance(args(i))) {
+        logInfo(s"arg $i not match: type $parameterWrapperType != ${args(i).getClass()}")
```
shivaram: This is very useful to have. Could we add a little bit more detail here to say what is expected and what was provided ?
i.e. Something like 
`logInfo(s"arg $i does not match: Expected type $parameterWrapperType, but got ${args(i).getClass()}")`

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
```diff
@@ -1,5 +1,7 @@
 package edu.berkeley.cs.amplab.sparkr
 
+import org.apache.spark.Logging
```
shivaram: The Logging API in Spark is marked as 'do not use outside of Spark` -- so lets stick to `println` for now and we can switch to the logger when we prepare for the merge

pkg/inst/tests/test_sparkSQL.R
```diff
@@ -294,6 +294,31 @@ test_that("column functions", {
   c4 <- approxCountDistinct(c) + countDistinct(c) + cast(c, "string")
 })
 
+test_that("group by", {
+  df <- jsonFile(sqlCtx, jsonPath)
+  df1 <- agg(df, name = "max", age = "sum")
+  expect_true(1 == count(df1))
+  df1 <- agg(df, age2 = max(df$age))
```
shivaram: This is very cool !

pkg/NAMESPACE
```diff
@@ -132,6 +134,9 @@ exportMethods("asc",
               "countDistinct",
               "sumDistinct")
 
+exportClasses("GroupedData")
+exportMethods("count", "agg", "avg")
```
shivaram: I think these methods are already exported by other classes like `RDD` or `Column` so we strictly don't need them here.

pkg/R/group.R
```diff
@@ -0,0 +1,74 @@
...
+
+setMethod("count",
+          signature(x = "GroupedData"),
+          function(x) {
+            dataFrame(callJMethod(x@sgd, "count"))
+          })
+
+#' Agg
+#'
+#' Aggregates on the entire DataFrame without groups.
```
shivaram: Actually in this case it'll be good to write some documentation here to say what are the forms of aggregation supported. i.e. say something like We support two forms of aggregation 

``` 
df2 <- agg(df, <column> = <aggFunction>)
example: df2 <- agg(df, age = "sum") # Clarify what column name will get created here ?
or
df2 <- agg(df, newColName = aggFunction(column)) 
example: df2 <- agg(df, ageSum = sum(df$age)) # Creates a new column named ageSum
```

shivaram:
Thanks for the update @davies -- this is looking very cool ! 

@cafreeman Will be good if you also took a look at the API here

pkg/NAMESPACE
cafreeman: `count` and `avg` definitely are, though I think `agg` still needs to be exported.

pkg/R/DataFrame.R
```diff
@@ -663,6 +663,33 @@ setMethod("toRDD",
...
+               sgd <- callJMethod(x@sdf, "groupBy", cols[[1]], listToSeq(cols[-1]))
+             } else {
+               jcol <- lapply(cols, function(c) { c@jc })
+               sgd <- callJMethod(x@sdf, "groupBy", listToSeq(jcol))
+             }
+             groupedData(sgd)
+           })
+
+
+setMethod("agg",
```
cafreeman: I think `agg` needs a generic here? I don't see it in existing `NAMESPACE` or as a method of base R

pkg/inst/tests/test_sparkSQL.R
cafreeman: +1. @davies this is awesome!

cafreeman:
Left a few inline comments, but overall LGTM. This is is great @davies and I'm looking forward to having this API in the package :)

pkg/R/DataFrame.R
davies: It's defined in group.R

cafreeman: Whoops, sorry about that!

pkg/R/group.R
davies: added

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
```diff
@@ -144,6 +144,7 @@ class SparkRBackendHandler(server: SparkRBackend) extends SimpleChannelInboundHa
       args: Array[java.lang.Object],
       parameterTypes: Array[Class[_]]): Boolean = {
     if (parameterTypes.length != numArgs) {
+      // println(s"num of arguments numArgs not match with ${parameterTypes.length}")
```
shivaram: Could we have this line back ? Actually the best thing to do might be to format this similar to the one below saying `Expect parameterTypes.length but got numArgs arguments`

davies: It's annoying  to see many logging about this, because it's normal to have multiple signature for a API with different number arguments. With `println`, we could not mute it, so I'd like to disable is by default.

shivaram: Ah I see - okay. I didn't realize this gets called for every method. In that case lets delete this line. The we should move the  println line to the top method to when `methods.isEmpty` is true to print `cannot find matching method $methodName. Candidates are $printMethods(methods)` or something like that. Does that sound better ?

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackendHandler.scala
```diff
@@ -100,11 +98,18 @@ class SparkRBackendHandler(server: SparkRBackend) extends SimpleChannelInboundHa
...
           matchMethod(numArgs, args, x.getParameterTypes)
-        }.head
-
-        val ret = selectedMethod.invoke(obj, args:_*)
+        }
+        if (methods.isEmpty) {
+          System.err.println(s"cannot find matching method ${cls.get}.$methodName. "
+            + s"Candidates are:")
+          selectedMethods.foreach { method =>
+            System.err.println(s"$methodName(${method.getParameterTypes})")
```
shivaram: very minor nit: Since getParameterTypes returns an array its better to do `mkString(",")` i.e.
`System.err.println(s"$methodName(${method.getParameterTypes.mkString(",")})")`

shivaram:
Thanks @davies - LGTM. Will merge when Travis passes

PR #181, opened 2015-02-25
Update method signatures to use generic arg

cafreeman:
Replace the `rdd` argument in all of the S4 methods with `x`. This will allow us to standardize the code as other Spark components get added and we need to set up multiple dispatch on S4 methods.

In any cases where `x` was used as a generic iterator, I've replaced it with `i` except in a few cases where a different letter made sense. For example, in some of the pair functions, we now use `function(k)` and `function(v)` for the key/value functions.

### DIFF
--- a/pkg/R/RDD.R
+++ b/pkg/R/RDD.R
@@ -265,7 +265,7 @@ setGeneric("unpersist", function(x) { standardGeneric("unpersist") })
 setMethod("unpersist",
           signature(x = "RDD"),
           function(x) {
-            callJMethod(getJRDD(x, "unpersist"))
+            callJMethod(getJRDD(x), "unpersist")
             x@env$isCached <- FALSE
             x
           })
@@ -399,7 +399,7 @@ setGeneric("collectAsMap", function(x) { standardGeneric("collectAsMap") })
 setMethod("collectAsMap",
           signature(x = "RDD"),
           function(x) {
-            pairList <- collect(rdd)
+            pairList <- collect(x)
             map <- new.env()
             lapply(pairList, function(i) { assign(as.character(i[[1]]), i[[2]], envir = map) })
             as.list(map)
--- a/pkg/R/RDD.R
+++ b/pkg/R/RDD.R
@@ -97,18 +97,18 @@ PipelinedRDD <- function(prev, func) {
 
 
 # The jrdd accessor function.
-setGeneric("getJRDD", function(x, ...) { standardGeneric("getJRDD") })
-setMethod("getJRDD", signature(x = "RDD"), function(x) x@jrdd )
-setMethod("getJRDD", signature(x = "PipelinedRDD"),
-          function(x, dataSerialization = TRUE) {
-            if (!is.null(x@env$jrdd_val)) {
-              return(x@env$jrdd_val)
+setGeneric("getJRDD", function(rdd, ...) { standardGeneric("getJRDD") })
+setMethod("getJRDD", signature(rdd = "RDD"), function(rdd) rdd@jrdd )
+setMethod("getJRDD", signature(rdd = "PipelinedRDD"),
+          function(rdd, dataSerialization = TRUE) {
+            if (!is.null(rdd@env$jrdd_val)) {
+              return(rdd@env$jrdd_val)
             }
 
             # TODO: This is to handle anonymous functions. Find out a
             # better way to do this.
             computeFunc <- function(split, part) {
-              x@func(split, part)
+              rdd@func(split, part)
             }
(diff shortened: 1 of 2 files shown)


pkg/R/RDD.R
```diff
@@ -97,18 +97,18 @@ PipelinedRDD <- function(prev, func) {
 
 
 # The jrdd accessor function.
-setGeneric("getJRDD", function(rdd, ...) { standardGeneric("getJRDD") })
-setMethod("getJRDD", signature(rdd = "RDD"), function(rdd) rdd@jrdd )
-setMethod("getJRDD", signature(rdd = "PipelinedRDD"),
-          function(rdd, dataSerialization = TRUE) {
-            if (!is.null(rdd@env$jrdd_val)) {
-              return(rdd@env$jrdd_val)
+setGeneric("getJRDD", function(x, ...) { standardGeneric("getJRDD") })
```
davies: This one only work for RDD, I'd like to keep it.

davies:
Should we change the signatures for all the methods, or for only for whose also works for DataFrame?

The downside is that we will make some of the code less readable.

cafreeman:
I considered both, but I think that having `x` for everything will ultimately be easier as the project grows and changes over time. It also provides a standard function format for everything in SparkR (e.g. your input object is always going to be `x`) and it conforms to existing R conventions (e.g. the signature for `lapply` in base R is `(X, FUN)` for all input types.

cafreeman:
It says the Travis CI build failed but all the unit tests pass on my machine. Looks like Travis might be down for maintenance?

pkg/R/RDD.R
shivaram: Yeah lets keep this as `rdd`

shivaram:
Yeah travis-ci seems to be down as seen from http://www.traviscistatus.com/ -- Hopefully it'll come back soon (lets give it a few hours)

shivaram:
Change otherwise LGTM -- I think its fine to keep RDD as `x` in all functions as it gives some uniformity across functions.

pkg/R/RDD.R
cafreeman: Good call, I didn't think about the inheritance issue.

Just switched it back to `rdd`

davies:
@cafreeman It makes sense. LGTM.

pkg/R/RDD.R
```diff
@@ -258,16 +258,16 @@ setMethod("persist",
...
 #' @aliases unpersist,RDD-method
 setMethod("unpersist",
-          signature(rdd = "RDD"),
-          function(rdd) {
-            callJMethod(getJRDD(rdd), "unpersist")
-            rdd@env$isCached <- FALSE
-            rdd
+          signature(x = "RDD"),
+          function(x) {
+            callJMethod(getJRDD(x, "unpersist"))
```
shivaram: This should be `callJMethod(getJRDD(x), "unpersist")` -- The bracket needs to move in

pkg/R/RDD.R
```diff
@@ -392,16 +392,16 @@ setMethod("collectPartition",
...
+setGeneric("collectAsMap", function(x) { standardGeneric("collectAsMap") })
 
 #' @rdname collect-methods
 #' @aliases collectAsMap,RDD-method
 setMethod("collectAsMap",
-          signature(rdd = "RDD"),
-          function(rdd) {
+          signature(x = "RDD"),
+          function(x) {
             pairList <- collect(rdd)
```
shivaram: This should be `collect(x)`

shivaram:
@cafreeman Travis is back and the latest test run showed a couple of failures. https://travis-ci.org/amplab-extras/SparkR-pkg/builds/52211253 -- I left comments inline that should hopefully fix them.

cafreeman:
Thanks @shivaram, I'll make the last few corrections first thing tomorrow.

This may be a silly question, but does Travis CI run extra tests or something? When I run all of the unit tests on my machine, everything passes, but your comments clearly highlight some areas where something is wrong. Is there something else I should be doing to catch these earlier?

shivaram:
Travis CI just runs `run-tests.sh` which is in our repo [1]. I'm not sure how you are running the tests but if you build SparkR using `./install-dev.sh` and then use `run-tests.sh` it should mimic Travis's behavior.

[1] https://github.com/amplab-extras/SparkR-pkg/blob/master/.travis.yml#L17

cafreeman:
That's odd. That's exactly what I've been doing.

cafreeman:
Fixed the typos and the tests pass again. Let's see if Travis CI agrees.

shivaram:
Thanks - LGTM. Merging this.

PR #182, opened 2015-02-26
Merge changes to master and update DataFrame methods to use new generic args

cafreeman:


shivaram:
Thanks @cafreeman . LGTM - I looked at the second commit which is change in DataFrame and that looks good.

PR #183, opened 2015-02-26
[SparkR 209] New DataFrame Methods

cafreeman:
Add several new methods to `DataFrame`:
- `first`
- `head`
- `cache`
- `persist`
- `unpersist`
- `schema`
- `dtypes`
- `columns`
- `names`

Also added a new `getStorageLevel` utility function so we don't duplicate code across persist methods.

Finally, this PR adds support for reading `StructType` and `StructField` directly into R via updates to `SerDe.scala` and `deserialize.R`


pkg/R/DataFrame.R
```diff
@@ -52,21 +53,108 @@ dataFrame <- function(sdf) {
...
+setGeneric("printSchema", function(x) { standardGeneric("printSchema") })
 
 setMethod("printSchema",
-          signature(df = "DataFrame"),
-          function(df) {
-            sdf <- df@sdf
-            schemaString <- callJMethod(sdf, "printSchema")
+          signature(x = "DataFrame"),
+          function(x) {
+            schemaString <- callJMethod(x@sdf, "printSchema")
```
davies: `printSchema` in JVM will write the scheme into stdout, and return nothing, so we should get the schema object, then call `treeString`
```
jschema <- callJMethod(x@sdf, "schema")
cat(callJMethod(jschema, "treeString"))
```

pkg/R/DataFrame.R
```diff
@@ -52,21 +53,108 @@ dataFrame <- function(sdf) {
...
             cat(schemaString)
           })
 
+#' Get schema object
+#' 
+#' Returns the schema of this DataFrame as a list of named lists. Each named
```
davies: Is it will be a wrapper of StructType in JVM ?

pkg/R/DataFrame.R
cafreeman: Are you sure? It works great right now. This is what my R console returns on a test DataFrame:

```R
> printSchema(df)
root
 |-- age: long (nullable = true)
 |-- name: string (nullable = true)
```

pkg/R/DataFrame.R
cafreeman: Right now, the backend intercepts a `StructType` and `StructField` and deserializes them automatically. I still need to add the ability write both types back out to the JVM.

pkg/R/DataFrame.R
```diff
@@ -193,11 +363,62 @@ setMethod("take",
...
+#' df <- jsonFile(sqlCtx, path)
+#' first(df)
+#' }
+
+setGeneric("first", function(x) {standardGeneric("first") })
+
+setMethod("first",
+          signature(x = "DataFrame"),
+          function(x) {
+            take(x, 1)
```
davies: Should we return an object instead of a list with one object?

pkg/R/DataFrame.R
```diff
@@ -52,21 +53,108 @@ dataFrame <- function(sdf) {
...
+#' path <- "path/to/file.json"
+#' df <- jsonFile(sqlCtx, path)
+#' columns(df)
+#'}
+setGeneric("columns", function(x) {standardGeneric("columns") })
+
+setMethod("columns",
+          signature(x = "DataFrame"),
+          function(x) {
+            sapply(schema(x), function(f) {
```
davies: Why not just call `callJMethod(x@sdf, "columns")`

pkg/R/DataFrame.R
```diff
@@ -52,21 +53,108 @@ dataFrame <- function(sdf) {
...
+#' dtypes(df)
+#'}
+
+setGeneric("dtypes", function(x) {standardGeneric("dtypes") })
+
+setMethod("dtypes",
+          signature(x = "DataFrame"),
+          function(x) {
+              lapply(schema(x), function(f) {
+                field <- c(f$name, f$dataType)
```
davies: The second part should be simpleString of DataType

pkg/R/DataFrame.R
```diff
@@ -52,21 +53,108 @@ dataFrame <- function(sdf) {
...
+          signature(x = "DataFrame"),
+          function(x) {
+            schemaString <- callJMethod(x@sdf, "printSchema")
             cat(schemaString)
           })
 
+#' Get schema object
+#' 
+#' Returns the schema of this DataFrame as a list of named lists. Each named
+#' list contains the elements of the StructField type, i.e. name, dataType, nullable.
```
davies: What's the benefits from have a list of fields instead of a wrapper of Java object (StructType object)?

The field could also be StructType (nested struct),

pkg/R/DataFrame.R
cafreeman: Hmmm, good question. Right now, `take` (and, by extension, `first`) call `collect` on the result, but `collect` returns a `data.frame` so it's not exactly a list, but it is a `data.frame` with only 1 row. It's not returning a formal `row` object like the standard API does, but returning a `data.frame` seemed like the most reasonable response in R since it gives you a clear and easily accessed object within R.

pkg/R/DataFrame.R
davies: When we test it locally, it works because the JVM and R process share the same console.

If we run those as separated process, then we can not see it in stdout of R.

pkg/R/DataFrame.R
davies: Sorry, I missed them in the begging. If we do it in this way, we will need to serialize all the DateType and re-construct all of them in R. We do it in this way in Python, but don't think it's a good idea in R, because OO is much harder in R than Python.

So, I'd like to have a wrapper for DataType in JVM, it will much easier.

pkg/R/DataFrame.R
cafreeman: Ahh, good to know. Thanks for pointing that out! I'll change it right now.

pkg/R/DataFrame.R
davies: Thanks for the explain, sounds reasonable.

When we call `first()`, I'd expect to have a object, then it's easy to access the column, for example:
```
obj <- df.first()
expect_true(obj$name == "AA")
expect_true(obj$age == 3)
```


We could keep it like this, then re-visit it later.

pkg/R/DataFrame.R
cafreeman: I believe that's what the existing backend is already doing with all the other data types, but if you know of a better/easier way I'm open to it.

Do you have an idea of what the wrapper might look like? Even just a rough sketch of what you have in mind would be useful.

pkg/R/DataFrame.R
cafreeman: You can do that now. Since `first` gives you a `data.frame` with named columns, we could run your example as it's written and it would work. Here's an example from the console:
```R
> obj <- first(df)
> class(obj)
[1] "data.frame"
> obj$name == "Michael"
[1] TRUE
```
Does that work?

davies: It works, thanks.

pkg/R/DataFrame.R
cafreeman: Good question. It's because I was mirroring the PySpark API version of `columns`, which grabs it out of the schema (as I'm sure you probably know :P )

The other benefit is that is doing it this way allows us to use `sapply` and return a `character` vector instead of a list. This allow us to still access all the elements inside the result by index location, but vectors print out more cleanly in the console than lists do (at least when all you've got is a series of strings).

pkg/R/DataFrame.R
davies: The DataType, I means ByteType(), StringType(), MapType(), not Int, or []String.

We can have this in schema():
```
function(x) { callJMethod(x$sdf, "schema") }
```
in dtypes:
```
function(x) {
  # handle Tuple SerDe
  callJMethod(x$sdf, "dtypes")
```

pkg/R/DataFrame.R
cafreeman: Right now it's actually returning `dataType.typeName` from the JVM. Is there a reason that `simpleString` is preferred to `typeName`?

pkg/R/DataFrame.R
davies: In PySpark, the schema() is on pure Python object, so it do in that way. There are hundreds of lines of codes to reconstruct the whole DataTypes.

It's easiy to turn a list in vector in R , if you want (in columns()).

pkg/R/DataFrame.R
davies: simpleString support nested structs, for example, MapType(keyType=StringType(), valueType=IntegerType()), you will got something like `map<string, int>`,

pkg/R/DataFrame.R
cafreeman: @shivaram I'm interested in your take on this as well. I know that PySpark handles all of the DataTypes with a series of classes and superclasses, which, as @davies points out, would be a lot harder to replicate in S4. I just assumed we were going to have to extend the backend to handle all of these going both ways, but if there's an easier way then great. I'm just not sure I follow what the wrapper function would look like in this case.

shivaram: @cafreeman I am not sure I completely understand the options here. I can see that PySpark implemented pure Python objects for every data type in SparkSQL and there is a lot of code for that which might be good to avoid.

The other way seems to me is to have a StructType S4 class in R which holds a `jobj` to a StructType class in JVM. (Similar to say what we do with DataFrame etc.) What are the methods we want to support on `StructType` ? Will it just be a `print`which can be used to print schemas ? I guess we might add methods to construct StructTypes in the future.
@davies is this what you had in mind ?

davies: Yes, thanks!

shivaram: Also @cafreeman we don't need to use S4 if the classes are really simple (see `jobj` for instance). In this case if all the class does is have `print` and some constructors S3 might be sufficient.

cafreeman: Ahhhh, it would essentially be a way to still get real values of certain types, but only when we want to as opposed to having the backend always deserialize it. So we could even just do some S3 functions that would handle the Spark types to get the info we need in different scenarios, right? Something that would end up looking like:
```R
schemaObj <- callJMethod(x@sdf, "schema")
out <- fields(schemaObj) # fields will call the appropriate schema methods and return only what we need
```
And we could have several different functions that mirror the methods of the Spark types that we need. Does that sound about right?

cafreeman: @shivaram just saw your second comment. Looks like we're on similar pages. I think that's a great suggestion @davies. I'll get that together tonight and update the PR.

cafreeman: @davies How about something like this:

```R
print.StructType <- function(st) {
  lapply(fields.StructType(st), function(i) {
    name <- callJMethod(i, "name")
    dataType <- callJMethod(i, "dataType")
    simpleString <- callJMethod(dataType, "simpleString")
    nullable <- callJMethod(i, "nullable")
    list("name" = name, "dataType" = simpleString, "nullable" = nullable)
  })
}

fields.StructType <- function(st) {
  callJMethod(st, "fields")
}
```

davies: It works to me, thanks!

pkg/R/DataFrame.R
cafreeman: Just commited `treeString` change

pkg/R/DataFrame.R
cafreeman: @davies Just committed wrapper functions for `structType` and `structField`. @shivaram your suggestion to look at `jobj.R` was a big help. I ended up using the same approach to basically replicate both types as well as their methods on Spark side, and they both have `print` methods in R as well. 

I think this approach is very extensible as well, and should come in handy as we need to include more of the Spark DataTypes.

pkg/R/DataFrame.R
cafreeman: Changed to `simpleString`

pkg/R/DataFrame.R
```diff
@@ -19,8 +19,9 @@ setClass("DataFrame",
          slots = list(env = "environment",
                       sdf = "jobj"))
 
-setMethod("initialize", "DataFrame", function(.Object, sdf) {
+setMethod("initialize", "DataFrame", function(.Object, sdf, isCached) {
   .Object@env <- new.env()
+  .Object@env$isCached <- isCached
```
shivaram: Just curious - do we use the `isCached` anywhere ?

pkg/R/DataFrame.R
```diff
@@ -52,21 +53,107 @@ dataFrame <- function(sdf) {
...
+#' df <- jsonFile(sqlCtx, path)
+#' dfSchema <- schema(df)
+#'}
+
+setGeneric("schema", function(x) { standardGeneric("schema") })
+
+setMethod("schema",
+          signature(x = "DataFrame"),
+          function(x) {
+            structType(callJMethod(x@sdf, "schema"))
```
shivaram: The doc above needs to be updated to say we return a `structType` object.

pkg/R/DataFrame.R
cafreeman: Yep, in `cache`, `persist`, and `unpersist`.

pkg/R/DataFrame.R
```diff
@@ -52,21 +53,107 @@ dataFrame <- function(sdf) {
...
+#' dtypes(df)
+#'}
+
+setGeneric("dtypes", function(x) {standardGeneric("dtypes") })
+
+setMethod("dtypes",
+          signature(x = "DataFrame"),
+          function(x) {
+            lapply(schema(x)$fields, function(f) {
+              field <- c(f$name, f$dataType.simpleString)
```
shivaram: This looks very neat !

pkg/R/DataFrame.R
```diff
@@ -193,11 +362,62 @@ setMethod("take",
...
+#' sqlCtx <- sparkRSQL.init(sc)
+#' path <- "path/to/file.json"
+#' df <- jsonFile(sqlCtx, path)
+#' head(df)
+#' }
+
+setMethod("head",
+          signature(x = "DataFrame"),
+          function(x, num = 6L) {
+          # Default num is 6L in keeping with R's data.frame convention
```
shivaram: Style nit - The `#` should needs space before to align with `take` on the next line

pkg/R/types.R
```diff
@@ -0,0 +1,47 @@
+# Utility functions for handling Spark DataTypes.
```
shivaram: I'd probably change this to SparkSQL types as I am not sure what other types we'll need to do this for.

pkg/R/types.R
```diff
@@ -0,0 +1,47 @@
...
+
+# Handler for StructType
+structType <- function(st) {
+  obj <- structure(new.env(parent = emptyenv()), class = "structType")
+  obj$jobj <- st
+  obj$fields <- lapply(SparkR:::callJMethod(st, "fields"), structField)
```
shivaram: This might be over-optimizing things, but it would be cool if these fields were populated lazily. i.e. only when the user wants to look at `fields` we actually make the `callJMethod`. One way to do that is to make these variables closures (i.e. `fields <- function() { lapply(SparkR:::callJMethod(st, "fields"), structField) }`) and users can call `st$fields()`  to get the fields.
@davies @concretevitamin Any thoughts ?

pkg/R/types.R
```diff
@@ -0,0 +1,47 @@
...
+
+#' Print a Spark StructType.
+#'
+#' This function prints the contents of a StructType returned from the
+#' SparkR JVM backend.
+#'
+#' @param x A StructType object
+#' @param ... further arguments passed to or from other methods
+print.structType <- function(x, ...) {
+  fieldsList <- lapply(x$fields, function(i) i$print)
```
shivaram: Minor style nit - I think we put `{}` even for one one line `function` definitions.

pkg/R/types.R
cafreeman: Ohhhh, that's what I wanted to do but I couldn't figure out a way to do it! That'd be perfect.

shivaram:
@cafreeman I just took a pass and this is looking pretty good. Lets see if @davies has any more comments and I will merge after that.

pkg/R/types.R
```diff
@@ -0,0 +1,47 @@
...
+#' Print a Spark StructField.
+#'
+#' This function prints the contents of a StructField returned from the
+#' SparkR JVM backend.
+#'
+#' @param x A StructField object
+#' @param ... further arguments passed to or from other methods
+print.structField <- function(x, ...) {
+  cat(x$print)
+}
```
davies: new line here.

pkg/NAMESPACE
```diff
@@ -97,3 +103,5 @@ export("jsonFile",
        "uncacheTable")
 
 export("sparkRSQL.init")
+export("print.structType",
```
davies: Is it possible to override the behavior of `str(structType)` without export these two methods?

For exported API, less is better,

pkg/R/types.R
```diff
@@ -0,0 +1,47 @@
...
+# Handler for StructField
+structField <- function(sf) {
+  obj <- structure(new.env(parent = emptyenv()), class = "structField")
+  obj$jobj <- sf
+  obj$name <- SparkR:::callJMethod(sf, "name")
+  obj$dataType <- SparkR:::callJMethod(sf, "dataType")
+  obj$dataType.toString <- SparkR:::callJMethod(obj$dataType, "toString")
+  obj$dataType.simpleString <- SparkR:::callJMethod(obj$dataType, "simpleString")
+  obj$nullable <- SparkR:::callJMethod(sf, "nullable")
+  obj$print <- paste("StructField(",
```
davies: I'd like to keep these attributes as less as possible, and call the Java method when needed, for example, dataType.simpleString, print, dataType.toString

pkg/NAMESPACE
cafreeman: As far as I know, if you want the print methods to trigger on those environments, you have to export the print method even if you don't export the actual wrapper function.

davies:
@cafreeman Once you could minimize the API and implementation of StructType, it looks good to me, thanks!

pkg/R/types.R
cafreeman: The problem is that since you can't string multiple method calls together in `callJMethod`, there isn't a good way to access the different values without having to just put more `callJMethod` calls in other functions. I think @shivaram's suggestion about making all of these lazy will allow us to have all of these attributes without it being too costly since none of them will actually execute when the `structField` is created.

pkg/NAMESPACE
davies: Did know R much, here is an example I took from shiny:
```
S3method(print,reactive)
S3method(print,shiny.appobj)
S3method(str,reactivevalues)
```

pkg/R/types.R
davies: lazy evaluation will be better.

pkg/R/types.R
concretevitamin: +1

pkg/R/types.R
concretevitamin: I think @shivaram's proposal is worth a shot (@cafreeman not sure if you have tried that). Also PipelinedRDD's getJRDD pattern is an example of how memoization could be done.

pkg/NAMESPACE
cafreeman: Those are also `namespace` exports though. They also have things like
```R
export(knit_print.html)
export(knit_print.shiny.appobj)
export(knit_print.shiny.render.function)
export(knit_print.shiny.tag)
export(knit_print.shiny.tag.list)
```
in their namespace for the same reason. Additionally, we have `print.jobj` in our namespace in order for the method to trigger correctly. I don't see where the issue is in this case.

pkg/R/DataFrame.R
cafreeman: Done.

pkg/R/DataFrame.R
```diff
@@ -52,21 +53,107 @@ dataFrame <- function(sdf) {
...
+#' @examples
+#'\dontrun{
+#' sc <- sparkR.init()
+#' sqlCtx <- sparkRSQL.init(sc)
+#' path <- "path/to/file.json"
+#' df <- jsonFile(sqlCtx, path)
+#' dtypes(df)
+#'}
+
+setGeneric("dtypes", function(x) {standardGeneric("dtypes") })
```
concretevitamin: nit: space after {

pkg/R/DataFrame.R
concretevitamin: We can get rid of the assignment here?

concretevitamin:
LGTM pending the lazy evaluation part.

cafreeman:
Just pushed the lazy evaluation changes and formatting updates.

concretevitamin:
This doesn't do memoization (evaluate once and cache the results) and is only delaying the calls - what do you all think about adding this optimization?

pkg/NAMESPACE
shivaram: AFAIK `S3method(print, structType)` is the same as `export(print.structType)` -- This is because the way S3 methods are dispatched is `method.type()` is called for `method(type)`

shivaram:
Yeah lazy evaluation doesn't give us memoization but I'm not sure latency costs are high enough to need memoization. We can come back to that later if we find its a problem (actually it'll be good to add some helper methods which automatically do memoization + lazy evaluation from JVM if this pattern comes up frequently)

shivaram:
New changes LGTM. And Travis is happy, so merging this

PR #184, opened 2015-02-26
[SPARKR-155] use socket in R worker

davies:
This PR will use socket to send and receive result from R worker. 

The reason we can not use stdin/stdout is that we plan to have daemon to launch the R worker,  then the stdin/stdout can not accessible from JVM.

### DIFF
--- a/pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/RRDD.scala
+++ b/pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/RRDD.scala
@@ -1,6 +1,7 @@
 package edu.berkeley.cs.amplab.sparkr
 
 import java.io._
+import java.net.{ServerSocket}
 import java.util.{Map => JMap}
 
 import scala.collection.JavaConversions._
@@ -11,6 +12,8 @@ import org.apache.spark.{SparkEnv, Partition, SparkException, TaskContext, Spark
 import org.apache.spark.api.java.{JavaSparkContext, JavaRDD, JavaPairRDD}
 import org.apache.spark.broadcast.Broadcast
 import org.apache.spark.rdd.RDD
+
+import scala.util.Try
 
 private abstract class BaseRRDD[T: ClassTag, U: ClassTag](
     parent: RDD[T],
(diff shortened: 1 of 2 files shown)


davies:
@shivaram Could you help to review this one?

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/RRDD.scala
```diff
@@ -27,21 +30,32 @@ private abstract class BaseRRDD[T: ClassTag, U: ClassTag](
 
   override def compute(split: Partition, context: TaskContext): Iterator[U] = {
 
-    val parentIterator = firstParent[T].iterator(split, context)
+    val serverSocket = new ServerSocket(0, 2)
```
shivaram: Any reason the backlog is set to 2 ? Will be good to add a comment

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/RRDD.scala
```diff
@@ -27,21 +30,32 @@ private abstract class BaseRRDD[T: ClassTag, U: ClassTag](
...
+    val listenPort = serverSocket.getLocalPort()
 
-    val pb = rWorkerProcessBuilder()
+    val pb = rWorkerProcessBuilder(listenPort)
+    pb.redirectErrorStream()  // redirect stderr into stdout
     val proc = pb.start()
+    val errThread =  startStdoutThread(proc)
 
-    val errThread = startStderrThread(proc)
+    // We use two socket ot separate input and output, then it's easy to manage
```
shivaram: `two sockets to separate` (spelling typo)

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/RRDD.scala
```diff
@@ -157,44 +164,33 @@ private abstract class BaseRRDD[T: ClassTag, U: ClassTag](
             dataOut.writeInt(1)
           }
 
-          for (elem <- iter) {
-            if (parentSerialized) {
+          if (parentSerialized) {
```
shivaram: any reason this loop was changed around (i.e. if before for) ?

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/RRDD.scala
```diff
@@ -27,21 +30,32 @@ private abstract class BaseRRDD[T: ClassTag, U: ClassTag](
...
 
-    val errThread = startStderrThread(proc)
+    // We use two socket ot separate input and output, then it's easy to manage
+    // the lifecycle of them to avoid deadlock.
+    // TODO: optimize it to use one socket
+
+    // the socket used to send out the input of task
+    serverSocket.setSoTimeout(10000)
+    val inSocket = serverSocket.accept()
+    val parentIterator = firstParent[T].iterator(split, context)
```
shivaram: So one reason the `parentIterator` call was at the beginning is that if two RRDDs were chained then the child R process's iterator will launch R do some computation and only when its over will the parent R process start.
This was mainly to avoid having too many R processes open at the same time -- Can we push this line back to the top of this method ?

shivaram:
Thanks @davies for the change. This is very useful and I think it'll improve performance in cases where the temp file was not in memory. I had some minor comments inline.

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/RRDD.scala
davies: It sounds reasonable to move it to the beginning, but I'm wondering that we still will have two R process.

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/RRDD.scala
```diff
@@ -12,6 +13,8 @@ import org.apache.spark.api.java.{JavaSparkContext, JavaRDD, JavaPairRDD}
 import org.apache.spark.broadcast.Broadcast
 import org.apache.spark.rdd.RDD
 
+import scala.util.Try
```
concretevitamin: group import

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/RRDD.scala
shivaram: I can't remember the details but @sun-rui investigated this in https://github.com/amplab-extras/SparkR-pkg/pull/87

shivaram:
Thanks @davies - LGTM but for the minor style comment @concretevitamin had (to put the scala import next to other scala imports).

shivaram:
Thanks @davies - Merging this.

PR #185, opened 2015-02-26
Unifies the implementation of lapply with lapplyParitionsWithIndex.

hlin09:


shivaram:
Thanks @hlin09 - LGTM. Merging this.

PR #186, opened 2015-02-27
[SPARKR-142][SPARKR-196] (Step 2) Replaces getDependencies() with cleanClosure to capture UDF closures and serialize them to worker.

hlin09:


shivaram:
@hlin09 Just checking, does this PR replace https://github.com/amplab-extras/SparkR-pkg/pull/158 ?

hlin09:
@shivaram. Sorry for the confusion, I planned to make this PR dependent on #158, but it now include all I did in #158. How could create dependencies between PRs? In short, this PR has all needed to fix the issue, and #158 includes only cleanClosure related in utils.R and test_utils.R.

shivaram:
Yeah its tricky to have dependent PRs. If this PR has all the changes, we can just use this one.

pkg/inst/worker/worker.R
```diff
@@ -42,11 +38,11 @@ for (pkg in packageNames) {
 }
 
 # read function dependencies
-depsLen <- SparkR:::readInt(inputCon)
-if (depsLen > 0) {
-  execFunctionDeps <- SparkR:::readRawLen(inputCon, depsLen)
-  # load the dependencies into current environment
-  load(rawConnection(execFunctionDeps, open='rb'))
+funcLen <- SparkR:::readInt(inputCon)
+if (funcLen > 0) {
```
shivaram: I think this will always be greater than 0 ? If not we don't have any function to compute so we should at least have `computeFunc <- identity` or something like that in the else clause

pkg/R/utils.R
```diff
@@ -288,3 +254,117 @@ convertEnvsToList <- function(keys, vals) {
...
+        # Continue to search in enclosure.
+        func.env <- parent.env(func.env)
+      }
+    }
+  }
+}
+
+# Utility function to get user defined function (UDF) dependencies (closure). 
+# More specifically, this function captures the values of free variables defined 
+# outside a UDF, and stores them in a new environment.
```
shivaram: In the last line `stores them in the function's environment` might be more accurate

pkg/R/utils.R
```diff
@@ -288,3 +254,117 @@ convertEnvsToList <- function(keys, vals) {
...
+      # up to global environment. There is no need to look into package environments
+      # above the global or namespace environment that is not SparkR below the global, 
+      # as they are assumed to be loaded on workers.
+      while (!identical(func.env, topEnv)) {
+        # Namespaces other than "SparkR" will not be searched.
+        if (!isNamespace(func.env) || getNamespaceName(func.env) == "SparkR") {
+          # Set parameter 'inherits' to FALSE since we do not need to search in
+          # attached package environments.
+          if (exists(nodeChar, envir = func.env, inherits = FALSE)) {
+            if (!isNamespace(func.env) ||
```
shivaram: Isn't this already ensured by the previous `if`. I am a bit confused as to why we need to check again.

pkg/R/utils.R
```diff
@@ -288,3 +254,117 @@ convertEnvsToList <- function(keys, vals) {
...
+             },
+             "<-" = {  # Assignment.
+               defVar <- node[[2]]
+               if (length(defVar) == 1 && typeof(defVar) == "symbol") {
+                 # Add the defined variable name into .defVars.
+                 assign(".defVars", 
+                        c(get(".defVars", envir = .sparkREnv), as.character(defVar)), 
+                        envir = .sparkREnv)
+               }
+               for (i in 3:nodeLen) {
```
shivaram: if `node[[2]]` was not a symbol, then we should call `processClosure` on that ?

pkg/R/utils.R
```diff
@@ -288,3 +254,117 @@ convertEnvsToList <- function(keys, vals) {
...
+    return
+  }
+  if (nodeLen > 1 && typeof(node) == "language") {
+    # Recursive case: current AST node is an internal node, check for its children. 
+    if (length(node[[1]]) > 1) {
+      for (i in 1:nodeLen) {
+        processClosure(node[[i]], oldEnv, argNames, newEnv)
+      }
+    } else {  # if node[[1]] is length of 1, might be an R primitive.
+      nodeChar <- as.character(node[[1]])
```
shivaram: Just to make sure I understand this correctly, the reason we care about `<-` and `function` is that these are statements that change the values of argNames and defVar.  In that case I think we should also switch on `=` as that is also an assignment operator.

I am not sure I completely get why we need to special case `{`. Its just a symbol and should work fine using the default case ?

pkg/R/utils.R
```diff
@@ -288,3 +254,117 @@ convertEnvsToList <- function(keys, vals) {
...
+      # as they are assumed to be loaded on workers.
+      while (!identical(func.env, topEnv)) {
+        # Namespaces other than "SparkR" will not be searched.
+        if (!isNamespace(func.env) || getNamespaceName(func.env) == "SparkR") {
+          # Set parameter 'inherits' to FALSE since we do not need to search in
+          # attached package environments.
+          if (exists(nodeChar, envir = func.env, inherits = FALSE)) {
+            if (!isNamespace(func.env) || 
+                  !nodeChar %in% getNamespaceExports("SparkR")) {  # Only include SparkR internals.
+              obj <- get(nodeChar, envir = func.env)
```
shivaram: its probably best to add `inherits=F` here too

pkg/inst/tests/test_utils.R
```diff
@@ -35,3 +35,50 @@ test_that("reserialize on RDD", {
...
+  expect_equal(actual, y)
+  actual <- get("g", envir = env)
+  expect_equal(actual, g)
+  
+  # Check for nested enclosures and package variables.
+  env2 <- new.env()
+  funcEnv <- new.env(parent = env2)
+  f <- function(x) { min(g(x) + y) }
+  environment(f) <- funcEnv  # enclosing relationship: f -> funcEnv -> env2 -> .GlobalEnv
+  newF <- SparkR:::cleanClosure(f)
```
shivaram: I think unit tests can call internal functions, so the `:::` might not be required.

shivaram:
Thanks @hlin09 for the change. The unit tests are especially very valuable ! I had some clarifications that I left inline.

pkg/inst/worker/worker.R
hlin09: Done.

pkg/R/utils.R
hlin09: Done.

pkg/R/utils.R
hlin09: I removed this check.

pkg/R/utils.R
hlin09: Yes we should.

pkg/R/utils.R
hlin09: You understand it correctly. I added check for `{` and `(` just to skip such special functions that does not need to go through the regular symbol search. It works fine if we remove it.

pkg/R/utils.R
hlin09: Done.

pkg/inst/tests/test_utils.R
hlin09: Done.

pkg/inst/tests/test_utils.R
```diff
@@ -35,3 +35,51 @@ test_that("reserialize on RDD", {
...
+  expect_equal(length(ls(env)), 2)  # y, g
+  actual <- get("y", envir = env)
+  expect_equal(actual, y)
+  actual <- get("g", envir = env)
+  expect_equal(actual, g)
+  
+  # Check for nested enclosures and package variables.
+  env2 <- new.env()
+  funcEnv <- new.env(parent = env2)
+  f <- function(x) { min(g(x) + y) }
```
concretevitamin: It's a little bit weird seeing `min()` applied to one argument - do we have a unary that we can use?

pkg/R/utils.R
```diff
@@ -288,3 +254,113 @@ convertEnvsToList <- function(keys, vals) {
...
+          !nodeChar %in% get(".defVars", envir = .sparkREnv)) {
+      func.env <- oldEnv
+      topEnv <- parent.env(.GlobalEnv)
+      # Search in function environment, and function's enclosing environments 
+      # up to global environment. There is no need to look into package environments
+      # above the global or namespace environment that is not SparkR below the global, 
+      # as they are assumed to be loaded on workers.
+      while (!identical(func.env, topEnv)) {
+        # Namespaces other than "SparkR" will not be searched.
+        if (!isNamespace(func.env) ||
```
concretevitamin: Perhaps add parenthesis here to make clear operator precedence.

pkg/inst/tests/test_utils.R
```diff
@@ -35,3 +35,51 @@ test_that("reserialize on RDD", {
   
   unlink(fileName)
 })
+
+test_that("cleanClosure on R functions", {
```
shivaram: Could we add a separate test case that mimics the bug in SPARKR-196 ? (i.e. using `t` as a variable and running lapply)

hlin09: Done.

pkg/R/utils.R
hlin09: Thanks. Done.

pkg/inst/tests/test_utils.R
hlin09: How about log?

pkg/inst/tests/test_utils.R
```diff
@@ -35,3 +35,64 @@ test_that("reserialize on RDD", {
...
+  t = 4  # Override base::t in .GlobalEnv.
+  f <- function(x) { x > t }
+  newF <- cleanClosure(f)
+  env <- environment(newF)
+  expect_equal(ls(env), "t")
+  expect_equal(get("t", envir = env), t)
+  actual <- collect(lapply(rdd, f))
+  expected <- as.list(c(rep(FALSE, 4), rep(TRUE, 6)))
+  expect_equal(actual, expected)
+})
```
shivaram: One last request for unit test: In the old code we were explicitly filtering out broadcast variables. I think that should automatically happen now but could we add a test case which looks something like this ?
```
a <- matrix(nrow=10, ncol=10, data=rnorm(100))
aBroadcast <- broadcast(sc, a)
normMultiply <- function(x) { norm(aBroadcast.value) * x }
# Check here that cleanClosure(normMultiply) only contains `aBroadcast` and not `a`
```

hlin09: Thanks. I added it.

shivaram:
Thanks @hlin09 -- LGTM.  Change looks pretty good to me. I'll see if @concretevitamin has any other comments before merging

One last curious thing: I found is that R actually supports assignment operators `->` and `->>` as well https://stat.ethz.ch/R-manual/R-patched/library/base/html/assignOps.html --  I haven't seen anybody use these and in fact they seem to be automatically converted to `<-` when I tried it out.

hlin09:
Thanks @shivaram and @concretevitamin for comments! Yes, I tried the same thing of "right assignment", which seems some syntax sugar that R provides. R actually does not define `->` primitives.

concretevitamin:
Also curious - @hlin09 do you have an idea of the reduction in typical # of bytes communicated due to this change? (Maybe there are some more interesting metrics but this is the one I can think of.)

hlin09:
Let's take a unit test ("several transformations on RDD (a benchmark on PipelinedRDD)" in  test_rdd.R) for example:
`nums <- 1:10
rdd <- parallelize(sc, nums, 2L)
rdd2 <- rdd
for (i in 1:6) {
  print(i)
  rdd2 <- lapplyPartitionsWithIndex(
    rdd2, function(split, part) {
      part <- as.list(unlist(part) * split + i)
    })
  print(object.size(rdd2))
}
rdd2 <- lapply(rdd2, function(x) x + x)
collect(rdd2)`

By using getDependencies(), we have # of bytes in each iteration: 
1: 4388 bytes
2: 10012 bytes
3: 19404 bytes
4: 44060 bytes
5: 255156 bytes
6: 1881652 bytes

By using this PR, we have # of bytes in each iteration:
1: 1884 bytes
2: 3492 bytes
3: 5100 bytes
4: 6708 bytes
5: 8316 bytes
6: 9924 bytes

concretevitamin:
:+1: LGTM

shivaram:
Nice test @hlin09  -- I also tested this with logistic regression locally and it brings down the size of dependencies from 267.5 KB to 86.3 KB (almost 3x reduction).

shivaram:
Merging this now. Thanks again for working on this change -- I think this fixes 3 or 4 issues on the JIRA

hlin09:
Thanks both for comments! @shivaram It is good to know it helps LR too.

PR #187, opened 2015-02-28
Three more DataFrame methods

cafreeman:
- `repartition`
- `distinct`
- `sampleDF`

`sampleDF` currently only uses the random seed version. It seems that in order to be able to set the seed manually, we'll need to be able to pass a Long type through `callJMethod` first. That can be expanded later unless someone already has it figured out.

### DIFF
--- a/pkg/R/DataFrame.R
+++ b/pkg/R/DataFrame.R
@@ -275,9 +275,9 @@ setGeneric("repartition", function(x, numPartitions) { standardGeneric("repartit
 #' @rdname repartition
 #' @export
 setMethod("repartition",
-          signature(x = "DataFrame", numPartitions = "integer"),
+          signature(x = "DataFrame", numPartitions = "numeric"),
           function(x, numPartitions) {
-            sdf <- callJMethod(x@sdf, "repartition", as.integer(numPartitions))
+            sdf <- callJMethod(x@sdf, "repartition", numToInt(numPartitions))
             dataFrame(sdf)     
           })
 
--- a/pkg/R/DataFrame.R
+++ b/pkg/R/DataFrame.R
@@ -335,7 +335,7 @@ setMethod("sampleDF",
           # TODO : Figure out how to send integer as java.lang.Long to JVM so
           # we can send seed as an argument through callJMethod
           signature(x = "DataFrame", withReplacement = "logical",
-                      fraction = "numeric"),
+                    fraction = "numeric"),
           function(x, withReplacement, fraction) {
             if (fraction < 0.0) stop(cat("Negative fraction value:", fraction))
             sdf <- callJMethod(x@sdf, "sample", withReplacement, fraction)
--- a/pkg/R/DataFrame.R
+++ b/pkg/R/DataFrame.R
@@ -253,6 +253,96 @@ setMethod("unpersist",
             x
           })
 
+#' Repartition
+#'
+#' Return a new DataFrame that has exactly numPartitions partitions.
+#'
+#' @param x A SparkSQL DataFrame
+#' @param numPartitions The number of partitions to use.
+#' @rdname repartition
+#' @export
+#' @examples
+#'\dontrun{
+#' sc <- sparkR.init()
+#' sqlCtx <- sparkRSQL.init(sc)
+#' path <- "path/to/file.json"
+#' df <- jsonFile(sqlCtx, path)
+#' newDF <- repartition(df, 2L)
+#'}
+
+setGeneric("repartition", function(x, numPartitions) { standardGeneric("repartition") })
+
+#' @rdname repartition
+#' @export
+setMethod("repartition",
+          signature(x = "DataFrame", numPartitions = "integer"),
+          function(x, numPartitions) {
+            sdf <- callJMethod(x@sdf, "repartition", as.integer(numPartitions))
+            dataFrame(sdf)     
+            })
+
+#' Distinct
+#'
+#' Return a new DataFrame containing the distinct rows in this DataFrame.
+#'
+#' @param x A SparkSQL DataFrame
+#' @rdname distinct
+#' @export
+#' @examples
+#'\dontrun{
+#' sc <- sparkR.init()
+#' sqlCtx <- sparkRSQL.init(sc)
+#' path <- "path/to/file.json"
+#' df <- jsonFile(sqlCtx, path)
+#' distinctDF <- distinct(df)
+#'}
+
+setMethod("distinct",
+          signature(x = "DataFrame"),
+          function(x) {
+            sdf <- callJMethod(x@sdf, "distinct")
+            dataFrame(sdf)
+            })
+
+#' SampleDF
+#'
+#' Return a sampled subset of this DataFrame using a random seed.
+#'
+#' @param x A SparkSQL DataFrame
+#' @param withReplacement Sampling with replacement or not
+#' @param fraction The (rough) sample target fraction
+#' @param seed Randomness seed value
+#' @rdname sampleDF
+#' @export
+#' @examples
+#'\dontrun{
+#' sc <- sparkR.init()
+#' sqlCtx <- sparkRSQL.init(sc)
+#' path <- "path/to/file.json"
+#' df <- jsonFile(sqlCtx, path)
+#' collect(sampleRDD(rdd, FALSE, 0.5, 1618L)) 
+#' collect(sampleRDD(rdd, TRUE, 0.5, 9L))
+#'}
+
+setGeneric("sampleDF",
+           function(x, withReplacement, fraction, seed) {
+             standardGeneric("sampleDF")
+          })
+
+#' @rdname sampleDF
+#' @export
+
+setMethod("sampleDF",
+          # TODO : Figure out how to send integer as java.lang.Long to JVM so
+          # we can send seed as an argument through callJMethod
+          signature(x = "DataFrame", withReplacement = "logical",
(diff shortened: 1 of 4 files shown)


pkg/R/DataFrame.R
```diff
@@ -253,6 +253,96 @@ setMethod("unpersist",
...
+#' df <- jsonFile(sqlCtx, path)
+#' newDF <- repartition(df, 2L)
+#'}
+
+setGeneric("repartition", function(x, numPartitions) { standardGeneric("repartition") })
+
+#' @rdname repartition
+#' @export
+setMethod("repartition",
+          signature(x = "DataFrame", numPartitions = "integer"),
```
shivaram: This is similar to the comment in #177 -- it might be better to allow numPartitions to be numeric and then convert it to integer. We could add a utils function for this as this is a common use case

pkg/R/DataFrame.R
```diff
@@ -253,6 +253,96 @@ setMethod("unpersist",
...
+          function(x, numPartitions) {
+            sdf <- callJMethod(x@sdf, "repartition", as.integer(numPartitions))
+            dataFrame(sdf)     
+            })
```
shivaram: Style nit: The `})` should be aligned with the beginning of `function`

pkg/R/DataFrame.R
```diff
@@ -253,6 +253,96 @@ setMethod("unpersist",
...
+#' df <- jsonFile(sqlCtx, path)
+#' distinctDF <- distinct(df)
+#'}
+
+setMethod("distinct",
+          signature(x = "DataFrame"),
+          function(x) {
+            sdf <- callJMethod(x@sdf, "distinct")
+            dataFrame(sdf)
+            })
```
shivaram: Same style nit as before

pkg/R/DataFrame.R
```diff
@@ -253,6 +253,96 @@ setMethod("unpersist",
...
+#' @param seed Randomness seed value
+#' @rdname sampleDF
+#' @export
+#' @examples
+#'\dontrun{
+#' sc <- sparkR.init()
+#' sqlCtx <- sparkRSQL.init(sc)
+#' path <- "path/to/file.json"
+#' df <- jsonFile(sqlCtx, path)
+#' collect(sampleRDD(rdd, FALSE, 0.5, 1618L))
```
shivaram: Since we don't support seed yet we can remove it from the documentation.

pkg/R/DataFrame.R
```diff
@@ -253,6 +253,96 @@ setMethod("unpersist",
...
+#' @export
+
+setMethod("sampleDF",
+          # TODO : Figure out how to send integer as java.lang.Long to JVM so
+          # we can send seed as an argument through callJMethod
+          signature(x = "DataFrame", withReplacement = "logical",
+                      fraction = "numeric"),
+          function(x, withReplacement, fraction) {
+            if (fraction < 0.0) stop(cat("Negative fraction value:", fraction))
+            sdf <- callJMethod(x@sdf, "sample", withReplacement, fraction)
```
shivaram: Actually this is one of those cases where having the Scala side accept an `Int` based version will be useful.

cc @davies @rxin -- Basically R doesn't have the concept of Long values, just ints or doubles. So its tricky to pass Long to a function.

pkg/R/DataFrame.R
cafreeman: I think that's a great idea. Do you happen to know how to grab the name of an incoming argument? I think the utility function could look like this:
```R
numToInt <- function(num) {
  if (!is.integer(num)) {
    warning(paste("Coercing", num, "to integer."))
    as.integer(num)
  }
}
```
The only problem is that `num` in the `warning()` function converts to the value, not the name, so we get:
```R
> numPartitions <- 5
> numToInt(numPartitions)
[1] 5
Warning message:
In numToInt(numPartitions) : Coercing 5 to integer.
```
I think it's ideal to have the message print the name so that we could use it as a true utility function and have the warning message adapt accordingly. This way the warning message will make sense relative to the function it's used in.

Ideally we want `Coercing numPartitions to integer.` in this case.

pkg/R/DataFrame.R
cafreeman: Fixed both. Sorry about that.

pkg/R/DataFrame.R
cafreeman: Actually, I think this will work:
```R
numToInt <- function(num) {
  if (!is.integer(num)) {
    warning(paste("Coercing", as.list(sys.call())[[2]], "to integer."))
    as.integer(num)
  }
}
```

shivaram: Can't we just use "numPartitions" as a string ? I mean we know the argument name when we write the function ? 
Also instead of `is.integer` we should check if `as.integer(num) == num` as we don't want to warn users if they pass `5` (we want to warn them if they pass 5.3)

cafreeman:
Just commited the `numToInt` function as well as the corresponding update to `repartition`.

pkg/R/DataFrame.R
cafreeman: I noticed that you seem to be able to call it with an `Int` in `spark-shell` but I'm guessing it's just doing an implicit conversion in the function call? I sure wasn't able to get it to work by passing an integer from R.

rxin: which variable do you need a long value?

shivaram: seed is long in the Scala API

rxin: would it work if you just pass an int?

shivaram: In the call to JVM we use the types of the arguments passed to figure out which function to call. So if we pass in `sample(..., Int)` it won't find any functions matching that definition in the JVM

pkg/R/DataFrame.R
cafreeman: We know the argument name in this specific case, but I can't imagine `numPartitions` is the only time we'll ever need to use this trick. Being able to write `5` instead of `5L` is convenient in any function where we need to send an integer, but there's no guarantee that those function arguments will always be named `numPartitions`. By using `sys.call`, we can use `numToInt` in any number of functions and know that the warning message will always be relevant to the actual function being called.

Good call on `as.integer`. I've changed that part.

pkg/R/utils.R
```diff
@@ -326,3 +326,11 @@ getStorageLevel <- function(newLevel = c("DISK_ONLY",
                          "OFF_HEAP" = SparkR:::callJStatic("org.apache.spark.storage.StorageLevel", "OFF_HEAP"))
 }
 
+# Utility function for functions where an argument needs to be integer but we want to allow
+# the user to type (for example) `5` instead of `5L` to avoid a confusing error message.
+numToInt <- function(num) {
+  if (!is.integer(num)) {
```
shivaram: This is still `if(!is.integer)` and not `if (as.integer(num) != num)` ?

pkg/R/utils.R
```diff
@@ -326,3 +326,11 @@ getStorageLevel <- function(newLevel = c("DISK_ONLY",
...
+    warning(paste("Coercing", as.list(sys.call())[[2]], "to integer."))
+    as.integer(num)
+  }
```
shivaram: We should return `as.integer(num)` outside the if

pkg/R/utils.R
cafreeman: I haven't pushed the change yet. :P

shivaram: Sorry I didn't catch that :) Also the sys.call() stuff is good.

cafreeman: Alright, NOW you can check. :)

It looks like this now:
```R
numToInt <- function(num) {
  if (as.integer(num) != num) {
    warning(paste("Coercing", as.list(sys.call())[[2]], "to integer."))
  }
  as.integer(num)
}
```

pkg/inst/tests/test_sparkSQL.R
```diff
@@ -197,4 +197,25 @@ test_that("head() and first() return the correct data", {
...
+  uniques <- distinct(df)
+  expect_true(inherits(uniques, "DataFrame"))
+  expect_true(count(uniques) == 3)
+})
+
+test_that("sampleDF on a DataFrame", {
+  df <- jsonFile(sqlCtx, jsonPath)
+  sampled <- sampleDF(df, FALSE, 1.0)
+  expect_equal(nrow(collect(sampled)), count(df))
+  expect_true(inherits(sampled, "DataFrame"))
```
shivaram: Can you also set the sampling fraction to something smaller (say 0.1) and test if count is less than 3 or 4 ? (It should be with very high probability)

pkg/R/DataFrame.R
```diff
@@ -253,6 +253,95 @@ setMethod("unpersist",
...
+          })
+
+#' @rdname sampleDF
+#' @export
+
+setMethod("sampleDF",
+          # TODO : Figure out how to send integer as java.lang.Long to JVM so
+          # we can send seed as an argument through callJMethod
+          signature(x = "DataFrame", withReplacement = "logical",
+                      fraction = "numeric"),
```
shivaram: very minor style nit -- the `fraction` here should align with `x`

shivaram:
Couple of minor comments - Otherwise LGTM.

cafreeman:
Fixed.

shivaram:
Thanks - LGTM. Merging this

PR #188, opened 2015-02-28
[SPARKR-189] [SPARKR-190] Column and expression

davies:
This PR introduce Column S4 class.

TODO: docs


davies:
cc @shivaram @cafreeman

pkg/R/column.R
```diff
@@ -0,0 +1,108 @@
...
+  .Object@jc <- jc
+  .Object
+})
+
+column <- function(jc) {
+  new("Column", jc)
+}
+
+# TODO: change Dsl to functions once update spark-sql
+col <- function(x) {
```
shivaram: I am not sure I understand this API -- what is the type of `x` here ? Is it a string ? If its a string, how do we know which dataframe's column needs to be selected ?

pkg/R/column.R
```diff
@@ -0,0 +1,108 @@
...
+  column(callJStatic("org.apache.spark.sql.Dsl", "col", x))
+}
+
+alias <- function(col, name) {
```
shivaram: alias, cast and also approxCountDistinct, countDistinct should become S4 functions

pkg/R/column.R
```diff
@@ -0,0 +1,108 @@
...
+#TODO(davies): like, rlike, startwith, substr, isNull, isNotNull, getField, getItem
+
+operators <- list(
+  "+" = "plus", "-" = "minus", "*" = "multiply", "/" = "divide", "%%" = "mod",
+  "==" = "equalTo", ">" = "gt", "<" = "lt", "!=" = "notEqual", "<=" = "leq", ">=" = "geq",
+  # we can not override `&&` and `||`, so use `&` and `|` instead
+  "&" = "and", "|" = "or"
+)
+
+createOperator <- function(op) {
```
shivaram: This is very cool !

pkg/inst/tests/test_sparkSQL.R
```diff
@@ -220,4 +220,31 @@ test_that("sampleDF on a DataFrame", {
...
+  df <- jsonFile(sqlCtx, jsonPath)
+  d <- collect(select(df, alias(df$age + 1, "age2")))
+  expect_true(names(d) == c("age2"))
+  df2 <- select(df, lower(df$name), abs(df$age))
+  expect_true(inherits(df2, "DataFrame"))
+  expect_true(count(df2) == 3)
+})
+
+test_that("column operators", {
+  c <- col("a")
```
shivaram: Can we use something like `df$newCol <- df$age + 5` as a test case ?  Actually this brings up the point that we'll need to allow a column to be assigned or added to a dataframe

davies: Right now, we can not do this in Scala/Python, because DataFrame in immutable, `withColumn()` will return a new DataFrame.

But we can try this in R, replace the underline `sdf` with new one.

shivaram: Yeah that will be cool to do. BTW I thought data frames were mutable even in Scala ? 
cc @rxin

pkg/R/column.R
davies: It's a string, this API is used for tests. All the columns are incomputable, just a expression.

davies:
@shivaram This PR block others (agg() and DataFrame APIs), so could we merge this in once it pass the tests? We could fill the docs and work on some details later.

shivaram:
@davies There are still errors due to cyclic namespace dependency in the build -- I haven't seen this error before, but from searching for the error message I found https://gist.github.com/anonymous/10018579 which might be helpful

cafreeman:
I'm looking for a case where there's an S3 function that would cause a collision with an S4 method but I haven't found anything yet. Anyone else?

shivaram:
I used the debugging mechanism listed at http://r.789695.n4.nabble.com/changes-in-recent-R-devel-revisions-td3330790.html -- And curiously the method that gives an error is `name` - Not sure why that would happen yet

cafreeman:
`name` or `names`?

My guess was that it was the fact that we're setting a generic for `alias` when it already exists in base R.

cafreeman:
Actually, a lot of the column methods create generics for methods that already exist (`max`, `min`, etc.). I know with `lapply` we have to just use the existing generic and create a new method that uses the same function definition. Could this be the problem?

pkg/inst/tests/test_sparkSQL.R
rxin: they are not mutable in the first cut. we can still make them mutable in 1.4, but we were worried it might confuse people since dataframes also have map/flatMap/reduce functions available ...

davies:
@shivaram I'm surprised that the variable name `name` caused the cyclic namespace dependency problem, it's fixed by changing it to another name.

shivaram:
yeah - so I don't know what the scoping rules are for code in package R files which are not inside any function. One thing we could do is to put all the generator code inside a function called generateColumnFunctions and just have one line in the R file which is not inside a function

BTW there is still a test error in approxCountDistinct as seen at https://travis-ci.org/amplab-extras/SparkR-pkg/builds/52796776

pkg/R/DataFrame.R
```diff
@@ -583,3 +583,29 @@ setMethod("foreachPartition",
...
+############################## DSL ##################################
+
+setMethod("$", signature(x = "DataFrame"),
+          function(x, name) {
+            column(callJMethod(x@sdf, "col", name))
+          })
+
+setGeneric("select", function(df, col, ...) { standardGeneric("select") } )
+
+setMethod("select", signature(df = "DataFrame", col = "character"),
```
cafreeman: For consistency, it should be `signature(x = "DataFrame")` in both methods

cafreeman:
I think there may be an issue with compiling the package. `DataFrame.R` uses the `column` class, but `column.R` comes after `DataFrame.R` in the `Collate` field.

However, if you switch the order so that `column.R` comes first, you get an error because there is a `first` method in `column.R` which is defined in `DataFrame.R`.

davies:
@cafreeman `first()` should be an API of RDD, then it will be fixed.

cafreeman:
I'll submit a PR real quick adding first to `RDD.R` and then we can just use `setMethod` in both the DataFrame and Column classes.

cafreeman:
Open here: https://github.com/amplab-extras/SparkR-pkg/pull/192/files

shivaram:
@davies @cafreeman is this up to date now ? Please ping this thread when its good to go.

cafreeman:
I think @davies just needs to revert the changes to `first` and we should be good.

pkg/R/DataFrame.R
```diff
@@ -583,3 +581,29 @@ setMethod("foreachPartition",
...
+
+setMethod("select", signature(x = "DataFrame", col = "character"),
+          function(x, col, ...) {
+            sdf <- callJMethod(x@sdf, "select", col, toSeq(...))
+            dataFrame(sdf)
+          })
+
+setMethod("select", signature(x = "DataFrame", col = "Column"),
+          function(x, col, ...) {
+            jcols <- lapply(list(col, ...), function(x) {
```
cafreeman: Can we change the argument to the inner function to something other than `x`? The scope works out fine, but it's confusing to have `x` be a `DataFrame` outside of the function while also being a `Column` inside the function.

Even something like
```R
function(c) {
  c@jc
}
```

davies:
@shivaram @cafreeman I think it's ready to go.

cafreeman:
LGTM. This is really cool @davies! I'm working the SQL-Like methods with new Column class right now :)

shivaram:
Alright - I'll merge this now to unblock other PRs. But lets leave the JIRA open as we need to add a few more operators and it'll be good to support things like `[[` to get columns out of a dataframe etc.

PR #189, opened 2015-03-01
Fix stderr redirection on executors

shivaram:
Calling `redirectErrorStream` without any arguments only tests whether stderr is redirected to stdout. JavaDoc at http://docs.oracle.com/javase/7/docs/api/java/lang/ProcessBuilder.html#redirectErrorStream() explains this

cc @davies


davies:
LGTM, thanks for the quick fix

PR #190, opened 2015-03-01
[SPARKR-79] Remove unused function

shivaram:
This is no longer needed with the new closure serialization strategy

cc @hlin09


hlin09:
Thanks for catch! LGTM.

PR #192, opened 2015-03-02
define generic for 'first' in RDD API

cafreeman:
Can now use this generic in `DataFrame` and `Column` methods.


pkg/R/RDD.R
```diff
@@ -841,6 +841,26 @@ setMethod("take",
...
+#' sc <- sparkR.init()
+#' rdd <- parallelize(sc, 1:10)
+#' first(rdd)
+#' }
+setGeneric("first", function(x) { standardGeneric("first") })
+
+setMethod("first",
+          signature(x = "RDD"),
+          function(x) {
+            take(x, 1)
```
davies: Should we return an object instead of list of one object.

cafreeman: Yep, you're right. Fixed.

cafreeman:
cc @shivaram

pkg/R/RDD.R
```diff
@@ -841,6 +841,26 @@ setMethod("take",
...
+#' sc <- sparkR.init()
+#' rdd <- parallelize(sc, 1:10)
+#' first(rdd)
+#' }
+setGeneric("first", function(x) { standardGeneric("first") })
+
+setMethod("first",
+          signature(x = "RDD"),
+          function(x) {
+            unlist(take(x, 1))
```
shivaram: Instead of `unlist` should we just use `take(x, 1)[[1]]` ?

cafreeman: We can. I'm not sure if it makes a difference either way? I'm good with either.

shivaram: `unlist` doesn't guarantee that we return one element (it just converts a list to a vector). Using `[[1]]` makes sure that only the first element is returned.

cafreeman: Fair enough. Just pushed the change.

shivaram:
LGTM. Will merge after travis passes.

cafreeman:
Thanks @shivaram! 

@davies Once this gets merged, we can update https://github.com/amplab-extras/SparkR-pkg/pull/188 to just rely on the RDD generic like you suggested and the collate issues should be good.

PR #193, opened 2015-03-03
[SPARKR-214] launch R worker by a daemon

davies:
The Spark executor could use lots of memory to cache some data, then it's very expensive to fork a R process (or easy to fail).

The solution is fork a lightweight daemon (in R) process in the beginning (when needed), call the daemon to launch R worker when needed.

This is only works in Linux/Mac, In Windows, it will still fork a R worker from JVM.

This also reduce the overhead to launch a task from about 350ms to 30ms (tested on Mac OS X 10.9)

TODO: improve failure handling. (delay it until we finalize the code for worker)

### DIFF
--- a/pkg/inst/worker/daemon.R
+++ b/pkg/inst/worker/daemon.R
@@ -1,7 +1,13 @@
 # Worker daemon
 
+# try to load `fork`, or install
+FORK_URI <- "http://cran.r-project.org/src/contrib/Archive/fork/fork_1.2.4.tar.gz"
+tryCatch(library(fork), error = function(err) {
+  install.packages(FORK_URI, repos = NULL, type = "source")
+})
+
 rLibDir <- Sys.getenv("SPARKR_RLIBDIR")
-script <- paste(rLibDir, "SparkR/worker/worker.R", sep="/")
+script <- paste(rLibDir, "SparkR/worker/worker.R", sep = "/")
 
 # preload SparkR package, speedup worker
 .libPaths(c(rLibDir, .libPaths()))
@@ -11,15 +17,23 @@ port <- as.integer(Sys.getenv("SPARKR_WORKER_PORT"))
 inputCon <- socketConnection(port = port, blocking = TRUE, open = "rb")
 
 while (TRUE) {
-  inport <- SparkR:::readInt(inputCon)
-  if (length(inport) != 1) {
-    quit(save="no")
+  ready <- socketSelect(list(inputCon), timeout = 1)
+  if (ready) {
+    inport <- SparkR:::readInt(inputCon)
+    if (length(inport) != 1) {
+      quit(save="no")
+    }
+    p <- fork::fork(NULL)
+    if (p == 0) {
+      close(inputCon)
+      Sys.setenv(SPARKR_WORKER_PORT = inport)
+      source(script)
+      fork::exit(0)
+    }
   }
-  p <- parallel:::mcfork()
-  if (inherits(p, "masterProcess")) {
-    close(inputCon)
-    Sys.setenv(SPARKR_WORKER_PORT = inport)
-    source(script)
-    parallel:::mcexit(0)
+  # cleanup all the child process
+  status <- fork::wait(0, nohang = TRUE)
+  while (status[1] > 0) {
+    status <- fork::wait(0, nohang = TRUE)
   }
 }
--- a/pkg/inst/worker/daemon.R
+++ b/pkg/inst/worker/daemon.R
@@ -14,9 +14,13 @@ while (TRUE) {
   ready <- socketSelect(list(inputCon))
   if (ready) {
     port <- SparkR:::readInt(inputCon)
+    # There is a small chance that it could be interrupted by signal, retry one time
     if (length(port) == 0) {
-      cat("quitting daemon", "\n")
-      quit(save = "no")
+      port <- SparkR:::readInt(inputCon)
+      if (length(port) == 0) {
+        cat("quitting daemon\n")
+        quit(save = "no")
+      }
     }
     p <- parallel:::mcfork()
     if (inherits(p, "masterProcess")) {
--- a/pkg/inst/worker/daemon.R
+++ b/pkg/inst/worker/daemon.R
@@ -10,48 +10,22 @@ suppressPackageStartupMessages(library(SparkR))
 port <- as.integer(Sys.getenv("SPARKR_WORKER_PORT"))
 inputCon <- socketConnection(port = port, open = "rb", blocking = TRUE, timeout = 3600)
 
-# Read from connection, retrying to read exactly 'size' bytes
-readBinFull <- function(con, size) {
-  # readBin() could be interruptted by signal, we have no way to tell
-  # if it's interruptted or the socket is closed, so we retry at most
-  # 20 times, to avoid the deadloop if the socket is closed
-  c <- 1
-  data <- readBin(con, raw(), size)
-  while (length(data) < size && c < 20) {
-    extra <- readBin(con, raw(), size - length(data))
-    data <- c(data, extra)
-    c <- c + 1
-  }
-  data
-}
-
-# Utility function to read the port with retry
-# Returns -1 if the socket was closed
-readPort <- function(con) {
-  data <- readBinFull(con, 4L)
-  if (length(data) != 4) {
-    -1
-  } else {
-    rc <- rawConnection(data)
-    ret <- SparkR:::readInt(rc)
-    close(rc)
-    ret
-  }
-}
-
 while (TRUE) {
-  port <- readPort(inputCon)
-  if (port < 0) {
-    cat("quitting daemon\n")
-    quit(save = "no")
-  }
-  p <- parallel:::mcfork()
-  if (inherits(p, "masterProcess")) {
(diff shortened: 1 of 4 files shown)


davies:
All the tests passed locally, how can we trigger the test in travis manually?

shivaram:
I jut re-triggered travis (looks like only those who have push access can re-trigger). If it fails again I'd actually suggest debugging with a Linux VM with only say 2G or 3G of memory. I think some of the race conditions etc. don't happen when on my Mac but happen in Travis VMs with limited memory.

Meanwhile @sun-rui -- could you take a look at this ?

shivaram:
@davies So I managed to debug the problem that existed with the `mcfork` approach you were using before. The problem is that `mcexit` doesn't exit the R process immediately but actually waits for the master process to either exit or send it a `SIGUSR1` [1] . So the reason the Travis build was failing is that 
these worker R processes would never exit and they'd use up memory over time.

One way to solve the problem might be to just send SIGUSR1 from the master to the child or I guess we could use something other than `mcexit`. I tried these but ran into some other problems with the socket connection that I didn't have time to investigate

[1] You can see the code for mcexit here https://github.com/wch/r-source/blob/R-3-1-branch/src/library/parallel/src/fork.c#L720

davies:
@shivaram Thanks to looking into it, I had switched to use fork package, then it works fine. The daemon process should wait the status from children, or they will become zombie process, but parallel package does not have wait() API, fork package fit to very well (but it's removed from package index, we should install it by link, or package it into SparkR?).

shivaram:
@davies I don't think its very good to depend on or use the fork package given that it is not actively maintained. And I did some more debugging yesterday and I fixed a couple of issues with using `mcfork` that made tests pass in my limited memory Ubuntu VM

The two main changes I made were:
- Setting `SIGUSR` inside the child process so that calling `mcexit` doesn't leave behind zombie processes. This also means that we don't need to wait for the child process in the daemon.
- Adding a version of `readInt` that retries till we have 4 bytes read. I noticed that while running the unit tests after `select` returned `TRUE`, sometimes I'd get a half-formed integer. Adding retry seems to fix that problem.

Both these changes are in #209 -- I've only opened it to see if it passed Travis. If it does we can port those changes here ?

davies:
@shivaram I had merged your changes here, and some changes, does it look good to you?

pkg/inst/worker/daemon.R
```diff
@@ -0,0 +1,57 @@
...
+
+port <- as.integer(Sys.getenv("SPARKR_WORKER_PORT"))
+inputCon <- socketConnection(port = port, open = "rb", blocking = TRUE, timeout = 3600)
+
+# Read from connection, retrying to read exactly 'size' bytes
+readBinFull <- function(con, size) {
+  # readBin() could be interruptted by signal, we have no way to tell
+  # if it's interruptted or the socket is closed, so we retry at most
+  # 20 times, to avoid the deadloop if the socket is closed
+  c <- 1
```
shivaram: I think this approach is fine, but I don't think the signal interruption concern is valid.  For example you can see in the R source code [1] that when the return code is `< 0` R raises an error (we could probably handle that with a tryCatch ?)

[1] https://github.com/wch/r-source/blob/354e3abe9574f4f830cb983bd1f6720e6e05e852/src/main/connections.c#L3985

davies: When the socket is opened in blocking mode, readBin() should read enough bytes util: 1) socket is closed; 2) timeout, We can verify these two, and it's not be interrupted when I `kill -CHLD PID` manually, you are right. 

I ran into issues that readInt() in daemon will return 0 (length is 0) when the child process die, after that readInt() will block until it got 4 bytes. What's the reason readBin() will return without enough bytes? (the socket is still open, the timeout is big enough).

shivaram: Yeah I don't know why exactly readBin behaves the way it does -- I think what happens is that calling `read` on a socket is best effort and returns how many ever bytes are available[2]. In fact I can't find any code in R [1] which ensures that the read is retried until all the required bytes are read. I think we don't run into this for larger message sizes, but in unit tests with only 4 bytes being sent very frequently we tend to run into this. 

I'm actually thinking this patterns needs to become more widespread (i.e. for the SparkR JVM backend as well ?)

[1] https://github.com/wch/r-source/blob/trunk/src/main/connections.c#L3797
[2] http://beej.us/guide/bgnet/output/html/multipage/recvman.html

shivaram:
BTW @davies I think I'm fine with merging the current approach but I am wondering if we should put this behind a flag (default can be true). This is a change that is hard to test on various machines / arch etc. so it'll be good to have a way to turn it off if required.

We could just use a special SparkConf variable (like `spark.sparkr.use.daemon`) or use JVM system properties

davies:
@shivaram will do.

davies:
@shivaram We heavy depend on `readBin` in deserialize.R, it's scaring that it can not read the fully expected bytes, so I'd like investigate more on it.

Here is the strace result of daemon process on Linux (some unrelated output were removed):
```
select(5, [4], [], NULL, {3600, 0}) = 1 (in [4], left {3338, 73798}) <261.926293>
recvfrom(4, "\0", 4096, 0, NULL, NULL) = 1 <0.000145>
select(5, [4], [], NULL, {3600, 0}) = 1 (in [4], left {3599, 999957}) <0.000179>
recvfrom(4, "\0", 4096, 0, NULL, NULL) = 1 <0.000164>
select(5, [4], [], NULL, {3600, 0}) = 1 (in [4], left {3599, 999998}) <0.000224>
recvfrom(4, "\261\374", 4096, 0, NULL, NULL) = 2 <0.000377>
select(5, [4], [], NULL, {3600, 0}) = ? ERESTARTNOHAND (To be restarted if no handler) <0.078264>
 --- SIGCHLD {si_signo=SIGCHLD, si_code=CLD_EXITED, si_pid=4792, si_status=0, si_utime=5, si_stime=2} ---
wait4(4792, [{WIFEXITED(s) && WEXITSTATUS(s) == 0}], WNOHANG, NULL) = 4792 <0.000055>
write(1, "quitting daemon", 15) = 15 <0.000090>
select(5, [4], [], NULL, {3600, 0}
```

The above is the behavior of the following code:
```
 port <- readInt(con)
if (length(port) == 0) {
  cat("quitting daemon")
  port <- readInt(con)
}
...
```

When we call `readInt()`, it actually call `select` and `recvfrom`.  It will handle the partial read from `recvfrom` and retry, until get enough number of bytes. 

 The `select` will be interrupted by SIGCHLD, then `readInt` return with length zero, then we saw `quitting daemon`.

The root cause is that it dose not handle signal correctly during `select`, so we need to handle it manually by
```
ready <- socketSelect(list(con))
if (ready) {
  port <- readInt()
}
```

even with this, it still will call select(), and return immediately, then it have a very small chance that get interrupted by signal. 

Also, in order to reduce the number of recvfrom, we should use buffered streaming in JVM.

For readBin() in deserialize.R, it's possible that select will be interrupted by signal, we may also need to do this (use socketSelect() before readBin()).

shivaram:
@davies Thanks a lot for digging into this. One idea I had is can we disable the `SIGCHILD` signal for the entire process ? Could you see if something like `signal("SIGCHLD","ignore")` in the daemon process helps ? http://svitsrv25.epfl.ch/R-doc/library/fork/html/signal.html

davies:
@shivaram I tried `signal("SIGCHLD","ignore")`, it did not help much, readInt() still return with zero length sometimes.

Do you think the current approach is good enough?

shivaram:
Thanks for the update. I like the idea of retry -- Let me try this out later tonight and get back

shivaram:
So I dug into this more last night and here is what I found

- When the select call in R gets a signal [it returns -1](https://github.com/wch/r-source/blob/R-3-1-branch/src/unix/sys-std.c#L157). 
- In the case of socketSelect the [return value is ignored](https://github.com/wch/r-source/blob/354e3abe9574f4f830cb983bd1f6720e6e05e852/src/main/connections.c#L5757) and only the FDs which are ready are picked up -- so this correctly handles signals (i.e. we can just go and call select again).
- But in the case of blocking socket read the loop checks the return value and [breaks](https://github.com/wch/r-source/blob/fbf5cdf29d923395b537a9893f46af1aa75e38f3/src/modules/internet/Rsock.c#L523) if the return value of select is not zero. 

This explains why if we get a `SIGCHLD`during the first select we are fine, but if we get one during the second select then we run into a problem.  I am not sure there is a nice solution around this though.

davies:
Yes, I saw the same. Retry is the only way we can do with readBin(), also we should add these for functions in deserialize.R, we can left that in another PR.

shivaram:
Yeah - I think the reason we don't run into this problem in the SparkR JVM communication is that we don't get signals to the process. We can just replace `readBin` with `readBinFully` or something in a future PR. Can you file a JIRA for this ?

And I'll take one more look at this change today.

davies:
Created : https://sparkr.atlassian.net/browse/SPARKR-226

Agreed that this is not urgent.

shivaram:
LGTM. Tested with flag set to true and false on a Ubuntu VM. The unit tests run around 2x faster while using fork (2.2 mins vs. 4.2 mins)

PR #194, opened 2015-03-03
[SPARKR-210] Some APIs in SQLContext

davies:
tables() and tableNames() is only available with latest spark-1.3


shivaram:
Thanks @davies. This is pretty cool. Any reason you didn't add `registerDataFrameAsTable` ? That seems to be the only method missing from those listed in SPARKR-210. 

Also, a different question is after SPARKR-210 will we be missing any methods or is that list complete ?

davies:
@shivaram Because we already have `registerTempTable(df, tableName)`, then `registerDataFrameAsTable(sqlCtx, df, tableName)` will not be useful, so I did not added it.

load() and loadExternalTable() will be added in another PR,  createDataFrame() will be in single PR, this is the difficult one.

That's all of them.

pkg/R/SQLContext.R
```diff
@@ -92,6 +110,34 @@ table <- function(sqlCtx, tableName) {
   dataFrame(sdf) 
 }
 
+
+#' Tablesq
```
shivaram: Typo - extra `q` at the end

pkg/R/SQLContext.R
```diff
@@ -102,11 +148,6 @@ table <- function(sqlCtx, tableName) {
 #' @export
 #' @examples
 #'\dontrun{
-#' sc <- sparkR.init()
```
shivaram: I don't think these examples need to be deleted. Its good style to have complete examples for each example that the user can run

pkg/NAMESPACE
```diff
@@ -123,11 +123,15 @@ exportMethods("asc",
               "sumDistinct")
 
 export("jsonFile",
+       "jsonRDD",
```
shivaram: This is not related to this PR, but can we sort this alphabetically ?

pkg/inst/tests/test_sparkSQL.R
```diff
@@ -20,6 +20,34 @@ test_that("jsonFile() on a local file returns a DataFrame", {
...
+test_that("test cache, uncache and clearCache", {
+  df <- jsonFile(sqlCtx, jsonPath)
+  registerTempTable(df, "temp")
+  cacheTable(sqlCtx, "temp")
+  uncacheTable(sqlCtx, "temp")
+  clearCache(sqlCtx)
+})
+
+test_that("test tableNames and tables", {
+  expect_true(length(tableNames(sqlCtx)) == 0)
```
shivaram: Can we test this after registering a table ?

pkg/R/SQLContext.R
davies: Make sense, will revert.

pkg/inst/tests/test_sparkSQL.R
davies: Good idea!

shivaram:
Thanks for the clarification - I left some minor comments inline but otherwise LGTM.

For merging this we can either
(a) wait for the next RC
(b) change the build.sbt to 1.3.0-SNAPSHOT. I'll then I'll test locally and merge

davies:
I tried to change it to 1.3.0-SNAPSHOT, but failed with:
```
[warn] 	Note: Unresolved dependencies path:
[warn] 		org.apache.spark:spark-core_2.10:1.3.0-SNAPSHOT (/Users/davies/work/SparkR-pkg/pkg/src/build.sbt#L23-54)
[warn] 		  +- edu.berkeley.cs.amplab:sparkr_2.10:0.1
[warn] 		org.apache.spark:spark-sql_2.10:1.3.0-SNAPSHOT (/Users/davies/work/SparkR-pkg/pkg/src/build.sbt#L23-54)
[warn] 		  +- edu.berkeley.cs.amplab:sparkr_2.10:0.1
sbt.ResolveException: unresolved dependency: org.apache.spark#spark-core_2.10;1.3.0-SNAPSHOT: not found
unresolved dependency: org.apache.spark#spark-sql_2.10;1.3.0-SNAPSHOT: not found
```

shivaram:
Did you do `mvn install` or `sbt/sbt publish-local` from the Spark master branch on your laptop ? If it still doesn't work you can try to add the following to resolvers and that should do the trick 

```
"Local Maven Repository" at Path.userHome.asFile.toURI.toURL + ".m2/repository",
```

cafreeman:
I don't think the library path normally looks like this, does it?

org.apache.spark#spark-core_2.10;1.3.0-SNAPSHOT

The '#' and ';' are throwing me off. Did you by any chance miskey something when you were updating build.sbt

davies:
@shivaram I could build it locally, but how can we test it in travis after merging it?

shivaram:
Ah I misunderstood -- We can't test it in travis with snapshot builds. I'll just test it locally on my machine and then merge.

shivaram:
Also just to clarify since sparkr-sql is a development branch only developers working on this feature will be building it for now. Which is why I think its okay to depend on a snapshot build for a few days while we continue working on things.

shivaram:
A slightly different solution in case we really want to make Travis happy. In sbt we can depend on a jar from any public URL. So I could for example upload a 1.3.0-SNAPSHOT jar to S3 and we can change out dependency to 

```
  "org.apache.spark" % "spark-core_2.10" % "1.3.0-SNAPSHOT" from "https://s3-us-west-1.amazonaws.com/sparkr-dep-1.3/2.10/spark-core_2.10-1.3.0-SNAPSHOT.jar"
```

Would anybody prefer that ?

davies:
+1

cafreeman:
I think I would. It's be nice to all be on the same page and also know that our test results will be consistent. Right now some of my unit tests fail on the snapshot because they're written for an "old" version of 1.3 and some of the data types have since changed.

cafreeman:
@shivaram so is that Amazon link currently live? If so, is there also one for `spark-sql_2.10`?

shivaram:
Not yet -- am looking at it now.

shivaram:
Yeah so this is setup now. To use this you'll need to do the following -- In build.sbt change
```
-  val defaultSparkVersion = "1.3.0"
+  val defaultSparkVersion = "1.3.0-SNAPSHOT-sparkr"
```
and in resolvers add the S3 bucket 
```
 resolvers ++= Seq(
+  "SparkR 1.3 Snapshots" at "https://s3-us-west-2.amazonaws.com/sparkr-dep-1.3/release",
```

Can you test this and also send a PR with it if it works ? FWIW the build here is from Spark master branch at https://github.com/apache/spark/commit/2db6a853a53b4c25e35983bc489510abb8a73e1d

cafreeman:
Building now. I'll let you now.

cafreeman:
Actually, I'm working on a PR right now that already updates things to a new version (and fixes things like `Dsl` vs `functions`, etc). I'll submit that stuff together before submitting new methods.

shivaram:
Thanks @davies for updating this. @cafreeman any more comments ?

pkg/R/SQLContext.R
```diff
@@ -68,6 +99,7 @@ sql <- function(sqlCtx, sqlQuery) {
   dataFrame(sdf)
 }
 
+
```
cafreeman: nit: extra line

pkg/R/SQLContext.R
```diff
@@ -92,6 +124,56 @@ table <- function(sqlCtx, tableName) {
   dataFrame(sdf) 
 }
 
+
```
cafreeman: nit: extra line

cafreeman:
The code looks good, but @davies you may want to check for extra lines. I think you picked up a bunch of extras when you merged.

davies:
@cafreeman These extra lines may be out of this PR, I think it's good to have two space line between each API (we use one between comment and API).

shivaram:
Alright I'm merging this PR -- At some point we should revisit some of the style guidelines we are using in this project and (a) document it and (b) use an automatic style checker

I'll open a JIRA for that.

PR #195, opened 2015-03-03
New 1.3 repo and updates to `column.R`

cafreeman:
Updated column to use `functions` instead of `Dsl` in accordance with the new API changes.

Also created separate S4 generics for `asc` and `desc` since they're actual `Column` methods


shivaram:
Thanks @cafreeman -- great to see Travis fixed. LGTM. Merging this.

PR #196, opened 2015-03-03
[SPARKR-198] shutdown the JVM after R process die

davies:
Create a tcp connection between R and JVM as heartbeat, The JVM will shutdown once this connection is closed by R (closed by OS if R process die)


shivaram:
Is this same as SPARKR-198 ?

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackend.scala
```diff
@@ -85,13 +87,35 @@ object SparkRBackend {
...
+          // any un-catched exception will also shutdown JVM
+          val buf = new Array[Byte](1024)
+          // shutdown JVM if R does not connect back in 10 seconds
+          serverSocket.setSoTimeout(10000)
+          val inSocket = serverSocket.accept()
+          serverSocket.close()
+          // wait for the end of socket, closed if R process die
+          inSocket.getInputStream().read(buf)
+          sparkRBackend.close()
+          System.exit(0)
```
felixcheung: Should return non-0 exit code - this is an unexpected case right?

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackend.scala
```diff
@@ -85,13 +87,35 @@ object SparkRBackend {
     try {
       // bind to random port
       val boundPort = sparkRBackend.init()
+      val serverSocket = new ServerSocket(0, 1)
+      val listenPort = serverSocket.getLocalPort()
```
felixcheung: My take was that we should be only doing this when R launches JVM, and not JVM launches R (SparkRRunner case)

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackend.scala
```diff
@@ -85,13 +87,35 @@ object SparkRBackend {
...
+
+      // wait for the end of stdin, then exit
+      new Thread("wait for stdin") {
+        setDaemon(true)
+        override def run(): Unit = {
+          // any un-catched exception will also shutdown JVM
+          val buf = new Array[Byte](1024)
+          // shutdown JVM if R does not connect back in 10 seconds
+          serverSocket.setSoTimeout(10000)
+          val inSocket = serverSocket.accept()
```
felixcheung: Handle exceptions and close serverSocket, close sparkRBackend.

davies: The purpose of this thread is to shutdown the JVM, and any un-catched exceptions will also do this, so it's not necessary to that complexity, I'd like to leave it as simple as possible.

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackend.scala
davies: good point, but I think that no one care about this return code.

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackend.scala
```diff
@@ -85,13 +87,35 @@ object SparkRBackend {
...
+          serverSocket.close()
+          // wait for the end of socket, closed if R process die
+          inSocket.getInputStream().read(buf)
```
felixcheung: Since it never sends anything read will block for the 1st byte and for the normal shutdown case it will throw IOException as the socket is closed? I'm not familiar with the general policy around these things but should this handle the expected exception and exit cleanly?

Also is there a race condition or this thread blocks indefinitely since another thread could be calling sparkRBackend.close() and this thread, this socket is never closed (I guess when the R process terminates it will be)

davies: In case of socket is closed, the `read()` will return -1 without IOException. 

`sparkRBackend.close()` could be called in main thread, then it exit. Because this thread is daemoned, so it will not block main thread from exiting.

Once the JVM exited, OS will cleanup all the resources (sockets, threads).

felixcheung: I think you might get an IOException if there is a socket error or something.
http://download.java.net/jdk7/archive/b123/docs/api/java/io/InputStream.html#read(byte[])

Woudn't that leave the R process zombied if sparkRBackend.close() is not called?

davies: there is an attempt to read at least one byte. If no byte is available because the stream is at the end of the file, the value -1 is returned

`sparkRBackend` is not R process, it's netty server. The thread will terminate (return -1 from read() or an exception during accept()) only when R process die first.

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackend.scala
felixcheung: I'm not sure about that. If this run in the cluster via SparkRRunner and R process dies, we would want to have a failure exit code to the Spark so that we would know there is an error with worker/executor.

davies:
@felixcheung Maybe there is a misunderstood that this patch is only used when R launch a JVM. If R process is launched by SparkRRunner, it will not go through this patch.

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackend.scala
```diff
@@ -85,13 +87,35 @@ object SparkRBackend {
...
       val path = args(0)
       val f = new File(path + ".tmp")
       val dos = new DataOutputStream(new FileOutputStream(f))
       dos.writeInt(boundPort)
+      dos.writeInt(listenPort)
       dos.close()
       f.renameTo(new File(path))
+
+      // wait for the end of stdin, then exit
+      new Thread("wait for stdin") {
```
felixcheung: "wait for socket disconnect"?

felixcheung:
Got it. I was confused about the two main(). It's good.

sun-rui:
Theoretically, it is more robust for the backend to periodically poll the R client to see if the client is alive than passively waiting for the socket close, as there are a couple of reasons that the socket close can't be detected (as described @ http://stackoverflow.com/questions/2428804/tcp-socket-closed, for example, the FIN packet may be lost). However, I agree that current solution is practical because the R client and backend run on the same host, where it is very less likely that socket close can't be detected.

what I am thinking is that could we eliminate the additional socket port for detection of R process exit. Since we now support only one connection from the R client, is it possible for us to detect the event of child channel close on the netty child channel, and if detected, close the R backend?

pkg/R/sparkR.R
```diff
@@ -149,11 +149,15 @@ sparkR.init <- function(
     }
     f <- file(path, open='rb')
     sparkRBackendPort <- readInt(f)
+    monitorPort <- readInt(f)
     close(f)
     file.remove(path)
-    if (length(sparkRBackendPort) == 0) {
+    if (length(sparkRBackendPort) == 0 || sparkRBackendPort == 0 ||
+     length(monitorPort) == 0 || monitorPort == 0) {
```
shivaram: align `length(monitorPort)` along with `length(sparkRBackendPort)`

pkg/R/sparkR.R
```diff
@@ -149,11 +149,15 @@ sparkR.init <- function(
...
       stop("JVM failed to launch")
     }
+    # never close this socket, JVM is waiting for it
+    .sparkREnv$monitorF <- socketConnection(port = monitorPort)
```
shivaram: Should we clear this variable during `sparkR.stop` ? Also we have a bunch of logic in `sparkR.stop` now that tries to reconnect to the backend if a valid connection does not exist. Can we simplify that logic based on whether this connection is open ?

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackend.scala
shivaram: So one nice thing to do here would be to move the `sparkRBackend.close()` to a finally block -- That should make sure we always run this regardless of exceptions ?

felixcheung:
@sun-rui agreed on both. I started with the active periodic ping, though agreed that socket error would likely be good enough.
Also similar thought on reusing existing channel.

davies:
@sun-rui It's not easy to periodic ping to JVM, because R does not support threads.

It's possible that we could re-use the socket to backend to detect wether the R process had die or not. But I think use different socket for different purpose could simplify the implementation, also introduce the possibility that we could close old connection (after failure) to backend and create new one (clean up some dirty state).

pkg/R/sparkR.R
davies: Should we support start SparkContext and stop multiple times? If yes, we should not clear this variable during spakrR.stop(), or the connection will be GCed (close), the JVM will shutdown.

In Python, we could do that, especially in tests.

sun-rui:
@davies , OK, that makes sense.

@shivaram , One more thought is that after we add this logic, is the logic in sparkR.stop trying to re-connect the backend still necessary?

pkg/R/sparkR.R
shivaram: So the design we are using in SparkR is that the JVM backend is tied one-to-one with the SparkContext. So the expected behavior at `sparkR.stop` is to stop the JVM backend (This could be necessary for any reason: testing or even just reducing memory used in the system etc.)

In unit tests we never actually call `sparkR.stop` -- we just call `sparkR.init` and that reuses the SparkContext across tests. BTW I'll continue this comment in the main PR thread as it'll be easier to follow there.

shivaram:
Continuing from https://github.com/amplab-extras/SparkR-pkg/pull/196#discussion_r25835361
the behavior we need is to make sure we stop the JVM when `sparkR.stop` is called.

So I think we can get rid of the code in `sparkR.stop` which tries to call `stopBackend` and just call `close` on the monitoring socket connection we have. We could do this after calling `sc.stop` to keep things clean (I think sc.stop is required for flushing event logs and things like that).

The other thing we can remove from `sparkR.R` is the hook to the finalizer to call `sparkR.stop` on R exit. This is not needed as when R exits the monitoring socket should get automatically closed.  The one corner case which might be a problem is that if users don't explicitly call `sparkR.stop` then `sc.stop` might never get invoked. But I think that should be fine for now.

davies:
@shivaram There is a case that we can not shutdown JVM in sparkR.stop, the R process is launched by JVM, and JVM may be shared by many services (for example, DBC), so we should not shutdown JVM in sparkR.stop.

If we want to shutdown JVM in some cases, could we have another API (say sparkR.shutdown()) to do it?

shivaram:
Well yes, in case the R process was launched by the JVM we shouldn't shutdown the JVM. But in your current implementation we won't have a monitoring socket at all right ? So it should be fine to check if the monitor socket is open and if so close it ?

shivaram:
Actually just to rephrase my earlier comment, our contract should be that `sparkR.stop` does the reverse of `sparkR.init` -- i.e. only if we launched a JVM in init we shut it down in stop

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackend.scala
davies: This patch only works in the case that JVM is launched by R, also the JVM process is an daemon process, the parent PID is 1, so no one will get the return code of it.

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackend.scala
davies: I think it's not necessary, we're going to shutdown the JVM, all the resources will be cleaned up automatically by OS.

pkg/R/sparkR.R
```diff
@@ -149,11 +156,14 @@ sparkR.init <- function(
...
     sparkRBackendPort <- readInt(f)
+    monitorPort <- readInt(f)
     close(f)
     file.remove(path)
-    if (length(sparkRBackendPort) == 0) {
+    if (length(sparkRBackendPort) == 0 || sparkRBackendPort == 0 ||
+        length(monitorPort) == 0 || monitorPort == 0) {
       stop("JVM failed to launch")
     }
+    assign(".monitorConn", socketConnection(port = monitorPort), envir = .sparkREnv)
```
shivaram: In addition to this, a couple of things can be removed now:
- Line 217 (`reg.finalizer`) can be removed as we don't need `sparkR.stop` to be run when the R process exits
- Line 28-37 can be refactored as follows: We check if a valid connection exists and if it doesn't we can just exit saying no connection to backend found (no need to reconnect). Also then we don't the `if` in Line 40 and the whole block can just go inside `if (connExists)`

@sun-rui does this sound good ?

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackend.scala
shivaram: The problem is not about resource cleanup but that `sparkRBackend.close()` may never get called. Here is a simple example I wrote up to exhibit the problem - https://gist.github.com/shivaram/8997ed1947bc19d4c933

If you run that gist you will see that exit is never called from the thread as the thread doesn't catch the IOException

pkg/R/sparkR.R
davies: I think we should still have reg.finalizer to call sparkR.stop(), it's better to shutdown SparkContext gracefully (call SparkContext.stop()).

shivaram: I can see that point, but the problem is `reg.finalizer` doesn't work in all cases. For example if you shutdown R and choose to save the workspace (or if you exit RStudio) then the socket connections get closed before `reg.finalizer` is run. So in those cases the JVM will shutdown before we get a chance to call `sc.stop`.  

I think the right thing to do might be to update our documentation and educate users that they should call `sparkR.stop` at the end of their program

davies: Make sense, will do.

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackend.scala
davies: The purpose of this thread is to shutdown JVM, it could be done by 1) reach the end of connection, call System.exit(0); 2) raise an exception in accept(), this un-catched exception will shutdown JVM without calling System.exit().

What's the purpose that we MUST call `sparkRBackend.close()` and `System.exit()` ?

shivaram: In (2) the problem is that uncaught exceptions in a daemon thread do not shutdown the JVM. Again the example I pasted above should show that.

davies:
@shivaram I tried to have a test to start and stop SparkContext multiple times, it did not work (it's related to managed the lifecycle of jobj), I will create a JIRA for that.

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SparkRBackend.scala
davies: Thanks, fixed.

pkg/R/sparkR.R
```diff
@@ -47,14 +34,16 @@ sparkR.stop <- function(env = .sparkREnv) {
   
     callJStatic("SparkRHandler", "stopBackend")
     # Also close the connection and remove it from our env
-    conn <- get(".sparkRCon", env)
+    conn <- get(".sparkRCon", envir = env)
     close(conn)
     rm(".sparkRCon", envir = env)
-    # Finally, sleep for 1 sec to let backend finish exiting.
-    # Without this we get port conflicts in RStudio when we try to 'Restart R'.
-    Sys.sleep(1)
```
shivaram: So it turns out we can't completely get rid of the finalizer because of this case where we try to 'Restart R' from RStudio. (You can reproduce this pretty easily. Just launch RStudio, load SparkR, call sparkR.init() and then press 'Restart R' in the Session menu -- Without the sleep the process will hang).

So the question is do we add a new finalizer which just calls `Sys.sleep(1)` or do we reuse `sparkR.stop` as the finalizer ? I'm in favor of the former option just to keep the finalizer simple.

shivaram:
Thanks @davies -- I just tested this locally with RStudio, command line and all the scenarios worked fine. Merging this.

PR #199, opened 2015-03-04
[SPARKR-203] load and save DataFrame

davies:
This PR also include HiveContext (but commented, depends on hive in spark-sql)


felixcheung:
Cool. The thing with HiveContext is that there is build of Spark without Hive support. I'm not sure there is a way to detect that. Maybe that should be a build profile.

davies:
@felixcheung Yes, I didnot figure out how to deal with it. We could re-visit it later.

shivaram:
@davies -- Can't we write a Java function which checks if Spark has Hive bindings built in ? Either using reflection or by inspecting the class loader we should be able to figure this out ?

pkg/R/DataFrame.R
```diff
@@ -747,3 +747,115 @@ setMethod("join",
...
+#' }
+
+setGeneric("saveAsTable", function(df, tableName, source, mode, ...) {
+  standardGeneric("saveAsTable")
+})
+
+setMethod("saveAsTable",
+          signature(df = "DataFrame", tableName = 'character', source = 'character',
+                    mode = 'character'),
+          function(df, tableName, source=NULL, mode="append", ...){
```
shivaram: minor style: `source = NULL, mode = "append"` (here and above)

pkg/inst/tests/test_sparkSQL.R
```diff
@@ -1,4 +1,6 @@
 library(testthat)
+library(devtools)
```
shivaram: We are not using `devtools` anywhere ?

pkg/R/SQLContext.R
```diff
@@ -218,20 +218,74 @@ uncacheTable <- function(sqlCtx, tableName) {
...
+#' @return DataFrame
+#' @export
+#' @examples
+#'\dontrun{
+#' sc <- sparkR.init()
+#' sqlCtx <- sparkRSQL.init(sc)
+#' df <- sparkRSQL.createExternalTable(sqlCtx, "myjson", path="path/to/json", source="json")
+#' }
+
+createExternalTable <- function(sqlCtx, tableName, path=NULL, source=NULL, ...) {
```
shivaram: Sorry there are a few more `path=NULL` instead of `path = NULL` in this PR. Could you try to do a find and replace in the entire file ?

shivaram:
@davies - Can we do something like what we do in Spark's Scala REPL ? I'm pasting the function below which creates HiveContext if hive is available

```
  def createSQLContext(): SQLContext = {
    val name = "org.apache.spark.sql.hive.HiveContext"
    val loader = Utils.getContextOrSparkClassLoader
    try {
      sqlContext = loader.loadClass(name).getConstructor(classOf[SparkContext])
        .newInstance(sparkContext).asInstanceOf[SQLContext]
      logInfo("Created sql context (with Hive support)..")
    }
    catch {
      case cnf: java.lang.ClassNotFoundException =>
        sqlContext = new SQLContext(sparkContext)
        logInfo("Created sql context..")
    }
    sqlContext
  }
```

davies:
@shivaram I had fixed on R side.

shivaram:
The R side fix is good. LGTM.

shivaram:
@felixcheung Nice catch on the HiveContext name -- I will you send a PR to fix it

davies:
@felixcheung Thanks!

PR #201, opened 2015-03-04
[SPARKR-156] phase 1: implement zipWithUniqueId() of the RDD class.

sun-rui:


davies:
LGTM

pkg/R/RDD.R
```diff
@@ -1427,6 +1427,47 @@ setMethod("setName",
...
+#' @aliases zipWithUniqueId,RDD
+setMethod("zipWithUniqueId",
+          signature(x = "RDD"),
+          function(x) {
+            n <- numPartitions(x)
+
+            partitionFunc <- function(split, part) {
+              index <- 0
+              result <- vector("list", length(part))
+              for (item in part) {
```
shivaram: It might be better to use `sapply` or `lapply` here instead of `for`

sun-rui:
According to the comment, I use mapply() instead of for statement. Actually, I did think about using lapply() or mapply(), but had a concern that involving a sequence vector may consume more memory than using a for statement. Not sure if my concern makes sense.

shivaram:
Thanks @sun-rui. LGTM, Merging this

PR #202, opened 2015-03-04
[SparkR-209] `explain`, `toJSON`, `isLocal`, `showDF`, `saveAsParquetFile` DataFrame methods

cafreeman:


pkg/R/DataFrame.R
```diff
@@ -78,15 +75,96 @@ setMethod("printSchema",
...
+#' @export
+#' @examples
+#'\dontrun{
+#' sc <- sparkR.init()
+#' sqlCtx <- sparkRSQL.init(sc)
+#' path <- "path/to/file.json"
+#' df <- jsonFile(sqlCtx, path)
+#' showDF(df)
+#'}
+setGeneric("showDF", function(x,...) { standardGeneric("showDF") })
```
shivaram: Actually what are the thoughts about just overriding the generic `show` ? In that case the default print method for the object will call this. On the one hand it'll make for a good user experience but I can also see the downside of running a job everytime

pkg/R/DataFrame.R
```diff
@@ -281,6 +359,34 @@ setMethod("repartition",
             dataFrame(sdf)     
           })
 
+#' toJSON
+#'
+#' Convert the rows of a DataFrame into JSON objects and return a StringRRDD
```
shivaram: StringRRDD is an internal detail - lets call this `return an RDD where each element contains a JSON string`

pkg/R/DataFrame.R
cafreeman: I tried that actually, but scratched it because the generic is fixed to one argument, so we wouldn't be able to add numRows as an additional argument in the overloaded method.

pkg/R/DataFrame.R
```diff
@@ -78,15 +75,96 @@ setMethod("printSchema",
...
+#' explain(df, TRUE)
+#'}
+setGeneric("explain", function(x, ...) { standardGeneric("explain") })
+
+#' @rdname explain
+#' @export
+setMethod("explain",
+          signature(x = "DataFrame"),
+          function(x, extended = FALSE) {
+            callJMethod(x@sdf, "explain", extended)
```
davies: We should get the string back and call `cat` in R, see PySpark 
```
        if extended:
            print self._jdf.queryExecution().toString()
        else:
            print self._jdf.queryExecution().executedPlan().toString()
```

cafreeman: Good call, fixed.

shivaram:
Thanks @cafreeman LGTM. @davies Any other comments ?

davies:
LGTM

shivaram:
Merging this. @cafreeman It'll be great if you can update the JIRA with what is still left.

PR #207, opened 2015-03-06
Remove print statement in SparkRBackendHandler

shivaram:
This print statement is noisy for SQL methods which have multiple APIs
(like loadDF). We already have a better error message when no valid
methods are found

cc @davies


cafreeman:
This is a much-needed change! Good call.

davies:
LGTM

PR #208, opened 2015-03-06
[SPARKR-188] Add profiling of R execution on worker side

lythesia:
@shivaram @sun-rui 

Few things might need further modification:

1. not sure I've understood the definition of boot/init/compute/.. time exactly, we can change them later if need.
2. to support profiling for `StringRRDD`, I modified code related to `dataStream` as well as the wirte ways on R worker side. Since we need a flag to tell if following data belongs to profling or string data, I made string writing line-by-line (on worker side) instead of `writeLines`.


shivaram:
@davies Could you take a look at this ? The profiling code might help us get some data on how much the the daemon, socket changes help

davies:
Before looking into the details, there are two questions:

1) Should we divide the output of task into sections (by section id) instead of special length? then each section could be different formats, frame based or line based. 

2) Could we fetch the metrics from driver? It will be much useful to have time along in the same console (run a job ,then fetch the metrics). SparkR does not support Accumulator now, so we may need a special way to fetch them. We have some similar metrics in PySpark, it's not useful in distributed environment because it's only reachable in logging of workers.

If 2) is not easy, we can do it in a separate PR.

lythesia:
1. Does this output mean the output of RDD or logger? Currently (before investigating deep into frame/line related methods), it's easy to tell `RRDD`, `PairwiseRRDD` and `StringRRDD`, and we can specify different log format for each.

2. For now the metrics are fetched from R worker side (like pyspark). If you mean calculating metrics simply on driver side I think it's hard to do that, since the situation about R work runing inside worker thread is invisible to driver.

shivaram:
1. @lythesia I think @davies is refererring to the output of the RDD from the R process to the JVM -- the idea being that we can divide the output into multiple sections instead of using a length field

2. I think this is more about trying to fetch the metrics from the RRDD to the driver. As you said the metrics can't be collected at the driver but it will be good to show the aggregate metrics on the driver side (its hard for users to look at the executor logs files). 
@davies -- To do this can we just use accumulators from the SparkR JVM backend ? i.e. I'm wondering if we can write a wrapper Scala function that will create accumulators, pass it on to RRDD and have another function to query the accumulators.

pkg/inst/worker/worker.R
```diff
@@ -1,5 +1,24 @@
 # Worker class
 
+# Utility function to get current system time in milliseconds
+# this code snippet borrowed from [R.utils](cran.r-project.org/package=R.utils)
+currentTimeMillis <- function() {
```
sun-rui: Borrowing code from R.tuils may have license issue because it is LGPL-licensed. We'd better re-write the code?

shivaram: Yeah I'm not even sure we need this - Wouldn't proc.time()[3] be sufficient ?

lythesia: I make this similar with java's `System.currentTimeMillis` which returns current datetime in milliseconds but not the process elapsed time.
We may do a little modification on it to avoid the license issue.

shivaram: Well all the cases in this file we only need relative time (i.e. time difference) -- so there shouldn't be much of a difference if we use milliseconds since epoch vs. milliseconds in this process.

lythesia: I see, I will fix it.

lythesia:
BTW, I did not undertand "divide RDD output into sections", can you give an example? @davies

davies:
@lythesia We could split the output from R tasks into two sections: RESULT, and METRICS, then in the worker.R:

```
writeInt(con, RESULT)
.... # write the result out
writeInt(con, METRICS)
....
```

In JVM:
```
while(true) {
   val section = readInt()
   section match {
   case RESULT => ....
   case METRICS => ...
   case END =>
       break
   }
} 
```

Then we could have different format in each section.

lythesia:
@davies Is this https://github.com/lythesia/SparkR-pkg/blob/master/pkg/inst/worker/worker.R#L155 similar with that MERICS you specified? Also for output there's a positive flag(or length) like RESULT.
And in JVM result&metrics are read like this https://github.com/lythesia/SparkR-pkg/blob/master/pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/RRDD.scala#L202. Is that you mean we should accumulate the metrics (maybe exclude the boot time?) during `case METRICS` and do reporting in `case END` ?

davies:
It's tricky to use a negative length as flag for special type. Even we have another type field, we still need a special length to tell the end of output. so never mind, we can go with current approach. Delay the refactor later, if needed.

davies:
If it's not easy to collect the metrics back to driver, we can revisit it once we have accumulator in R.

shivaram:
@lythesia Any update on this ?

lythesia:
@shivaram rewrite time profiling; change metrics to seconds

shivaram:
Yeah it would be great if you could merge with current master

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/RRDD.scala
```diff
@@ -259,26 +259,47 @@ private class StringRRDD[T: ClassTag](
...
 
   lazy val asJavaRDD : JavaRDD[String] = JavaRDD.fromRDD(this)
 }
 
+private object SpecialLengths {
+  val END_OF_STREAM = 0
+  val TIMING_DATA   = -1
+}
+
+private class Logger extends Serializable {
```
shivaram: Can we name this as `TimeLogger` or `WorkerProfiler` or something like that ?

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/RRDD.scala
```diff
@@ -136,6 +141,9 @@ private abstract class BaseRRDD[T: ClassTag, U: ClassTag](
               dataOut.writeInt(elemArr.length)
               dataOut.write(elemArr, 0, elemArr.length)
             } else {
+              // write string(for StringRRDD)
+              val str = elem.asInstanceOf[String]
```
shivaram: I'm not sure why need this change -- we aren't using `str` anywhere ?

lythesia: sorry for ignoring nit

shivaram:
Thanks @lythesia -- So it looks like this change also changes the way we handle string output. While going from strings to binary is generally a good thing, the `readString`, `writeString` implementations in `SerDe` are limited to ASCII strings only. 

It may be that this doesn't cause a problem, but I was wondering could we avoid changing it in this PR ?

lythesia:
@shivaram I had specified the reason I made the StringRRDD output change to line-by-line. 

In orginal implemention, the R worker simply write all outputs in one `writeStrings`, and JVM only need to read strings from the inputstream (in actual, it read in [line-by-line](https://github.com/amplab-extras/SparkR-pkg/pull/208/files#diff-d30e0fa6bbb931a15b0e50796c8f0c5dL271)).

For profiling, we need to write the metrics to same inputstream and need a flag to diff them with string data. And to keep it consistent with `read` styles of other RDDs, I transfer it to binary.

Seeking for better solutions, too.

shivaram:
So one thing we could do is just use the stream in mixed mode ? For example in startStdInThread we write out some values using a DataOutputStream and then use a PrintWriter. Similarly we could first read the special lengths in binary format and after we reach the right number of bytes we could switch ?

lythesia:
Currently profiling&rdd data feed in the same socket channel, so we cannot open two inputstreams at same time.

I've considered the latter. I'll try to change into that way.

lythesia:
Met with some problems:
1. `startStdInThread ` does wirte binary&string to stream, but R worker reads them both in binary mode.
2. I tried using BufferedRead reads strings, then construct DataInputstream from same inputstream and read leads to an EOF exception. Also we cannot `mark` `reset` on both different streams.
3. Since the metrics are genrated after rdd output's done, we should read data first then metrics. If so, we also need flags or length to tell the boundary between data and metrics (for StringRRDD, we should know # of lines to be output, and do a dirty hack to record it in JVM each `read`), I'm not sure which is better comparing to current style.

shivaram:
Thanks @lythesia for the update

@davies do you have any thoughts on this ? I am a bit busy today but can take a more detailed look at this tomorrow

shivaram:
I think the right way to do this is to actually improve our `readString` and `writeString` methods to handle UTF-8. On the R side instead what we need to do is to use `nchar(type = "bytes")` in writeString. This will return the number of bytes in the string and then writeBin should do the right thing. Then in the Scala side in `readString` we should get bytes and create a string with character encoding UTF-8 (`new String(byte[], "UTF-8")`).

FWIW in Python it looks like things are handled similarly to send strings from JVM to Python (See https://github.com/apache/spark/blob/540b2a4eabe0bad2455f5140c4ad6a315e37cc3f/core/src/main/scala/org/apache/spark/api/python/PythonRDD.scala#L588)

@davies @sun-rui Any thoughts on this ?

shivaram:
Just to add one more thing -- On the R side we should call `enc2utf8` on the strings before writing them out

davies:
@shivaram Agreed, we should do in this way.

lythesia:
@shivaram  Now write out in utf-8 on worker side. I wonder if we should change other string IO into utf-8 standard? Maybe it can be done in another pr?

shivaram:
Thanks @lythesia - I'll take a look at this soon. 

@davies Would be great if you could also take a look

pkg/R/serialize.R
```diff
@@ -152,5 +153,14 @@ writeArgs <- function(con, args) {
 }
 
 writeStrings <- function(con, stringList) {
-  writeLines(unlist(stringList), con)
+  # writeLines(unlist(stringList), con)
+  strs <- unlist(stringList)
```
shivaram: we dont need the unlist here anymore - you can just use `length(stringList)`

pkg/inst/worker/worker.R
```diff
@@ -1,5 +1,22 @@
...
+  as.numeric(Sys.time())
+}
+
+# Get elapsed time
+elapsedSecs <- function() {
+  proc.time()[3]
+}
+
+# Constants
+SpecialLengths <- list(END_OF_STERAM = 0L, TIMING_DATA = -1L)
```
shivaram: This should be `specialLengths`

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/RRDD.scala
```diff
@@ -24,10 +26,15 @@ private abstract class BaseRRDD[T: ClassTag, U: ClassTag](
...
     broadcastVars: Array[Broadcast[Object]])
   extends RDD[U](parent) {
+  protected var dataStream: DataInputStream = _
+  val logger: TimeLogger = new TimeLogger
   override def getPartitions = parent.partitions
 
   override def compute(split: Partition, context: TaskContext): Iterator[U] = {
 
+    // Timing start
+    logger.boot = System.currentTimeMillis / 1000
```
davies: Should we convert `System.currentTimeMillis` to double or use 1000.0 ?

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/RRDD.scala
```diff
@@ -269,26 +269,44 @@ private class StringRRDD[T: ClassTag](
...
+  var init: Double = _
+  var broadcast: Double = _
+  var input: Double = _
+  var compute: Double = _
+  var output: Double = _
+
+  def reportTime: String = {
+    ("Times: boot = %.1f s, init = %.1f s, broadcast = %.1f s, " +
+      "read-input = %.1f s, compute = %.1f s, write-output = %.1f s, " +
+      "total = %.1f s").format(
```
davies: This timings are usually a few ms, 1 digit is not enough for them, 3 or 4 will be better.

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/RRDD.scala
```diff
@@ -269,26 +269,44 @@ private class StringRRDD[T: ClassTag](
...
+        readStringBytes(dataStream, length)
+      case _ => null
     }
   }
 
   lazy val asJavaRDD : JavaRDD[String] = JavaRDD.fromRDD(this)
 }
 
+private object SpecialLengths {
+  val END_OF_STREAM = 0
```
davies: Maybe we will drop the last `\0` in string, then it's possible to have a zero-length string, should we use negative values for all special lengths?

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/RRDD.scala
```diff
@@ -161,9 +169,31 @@ private abstract class BaseRRDD[T: ClassTag, U: ClassTag](
...
+      length match {
+        case SpecialLengths.TIMING_DATA =>
+          // Timing data from R worker
+          logger.boot = dataStream.readDouble - logger.boot
+          logger.init = dataStream.readDouble
+          logger.broadcast = dataStream.readDouble
+          logger.input = dataStream.readDouble
+          logger.compute = dataStream.readDouble
+          logger.output = dataStream.readDouble
+          logInfo(logger.reportTime)
```
davies: We did not have any benefits from `TimeLogger`, I'd like to pull them here, it will be more readable.
```
val boot = dataStream.readDouble - bootTime
...
logInfo(xxx, boot, ...)
```

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/RRDD.scala
lythesia: actually it's currently not used, may I remove it?

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/RRDD.scala
lythesia: or we'll keep a `bootTime` as member of `BaseRRDD`. I made it for thought of possible future use (not sure if make sense).

davies:
LGTM, thanks!

pkg/inst/worker/worker.R
```diff
@@ -54,12 +77,21 @@ if (isEmpty != 0) {
...
     output <- computeFunc(splitIndex, data)
+    # Timing computing
+    computeElap <- elapsedSecs()
+
     if (serializer == "byte") {
       SparkR:::writeRawSerialize(outputCon, output)
     } else {
-      SparkR:::writeStrings(outputCon, output)
+      # write lines one-by-one with flag
+      lapply(output, function(line) SparkR:::writeString(outputCon, line))
```
shivaram: Why don't we use `writeStrings` here ? Is it because we don't want to write the length of the list out ?

shivaram: Just to clarify, `writeStrings` isn't used anywhere else, so if we don't need it here, we can just remove the function.

shivaram:
@lythesia Sorry for the delay in looking at this. I had one minor comment, but otherwise the change looks good to me.

lythesia:
@shivaram Yes, we now do not use the length of string list and I've removed it.

shivaram:
Thanks @lythesia LGTM. Merging this and cherry-picking this to sparkr-sql. 

@lythesia Can you also open a PR against amplab-extras/spark ? If you have trouble porting this to the new code locations let me know and I can try to cherry-pick this appropriately.

davies:
@shivaram I think we should create a new PR for these new commits on spark-sql, then people do not need to go over the huge PR again and again. It's easy for me to pull the new changes (just two steps of  merge).

shivaram:
@davies Not sure I follow - Do you mean that you can cherry-pick these changes instead of opening PRs ?

davies:
@shivaram I can pull all the changes in sparkr-sql branch by merge:
```
git checkout remote_r # local branch with new directory
git pull amp sparkr-sql
git checkout R     # local branch in Spark,  could be used for PR
git merge remote_r
```

So we can hold on this, do a round of update in another PR.

PR #209, opened 2015-03-08
Launch R worker using a daemon.

shivaram:
Modification of patch by @davies in #193

### DIFF
--- a/pkg/inst/worker/daemon.R
+++ b/pkg/inst/worker/daemon.R
@@ -1,0 +1,70 @@
+# Worker daemon
+
+rLibDir <- Sys.getenv("SPARKR_RLIBDIR")
+script <- paste(rLibDir, "SparkR/worker/worker.R", sep="/")
+
+# preload SparkR package, speedup worker
+.libPaths(c(rLibDir, .libPaths()))
+suppressPackageStartupMessages(library(SparkR))
+
+port <- as.integer(Sys.getenv("SPARKR_WORKER_PORT"))
+
+inputCon <- socketConnection(host = "localhost",
+                             port = port,
+                             blocking = TRUE,
+                             open = "rb",
+                             timeout = 6000,
+                             server = FALSE)
+
+# Read from connection, retrying to read exactly 'size' bytes
+readWithRetry <- function(con, size) {
+  data <- readBin(con, raw(), size, endian = "big")
+  # Since this function is called after
+  # calling `select`, 0 bytes read implies socket is closed
+  if (length(data) == 0) {
+    return(data)
+  }
+  while (length(data) < size) {
+    bytesToRead = size - length(data)
+    extra <- readBin(con, raw(), bytesToRead, endian = "big")
+    data <- c(data, extra)
+  }
+  stopifnot(length(data) == size)
+  data
+}
+
+# Utility function to read the port with retry
+# Returns -1 if the socket was closed
+readIntPort <- function(con) {
+  data <- readWithRetry(con, 4L)
+  if (length(data) != 4) {
+    -1
+  } else {
+    rc <- rawConnection(data)
+    ret <- SparkR:::readInt(rc)
+    close(rc)
+    ret
+  }
+}
+
+while (TRUE) {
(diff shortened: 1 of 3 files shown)


pkg/inst/worker/daemon.R
```diff
@@ -0,0 +1,70 @@
...
+    if (inport < 0) {
+      cat("quitting daemon\n")
+      q(save="no")
+    } else {
+      #cat("got port ", inport, "\n")
+      p <- parallel:::mcfork()
+      if (inherits(p, "masterProcess")) {
+        close(inputCon)
+        # Set SIGUSR1 so that child can exit
+        tools::pskill(Sys.getpid(), tools::SIGUSR1)
```
davies: This looks hacky, could we just call `quit(save="no")` in the end?

pkg/inst/worker/daemon.R
```diff
@@ -0,0 +1,70 @@
...
+readWithRetry <- function(con, size) {
+  data <- readBin(con, raw(), size, endian = "big")
+  # Since this function is called after
+  # calling `select`, 0 bytes read implies socket is closed
+  if (length(data) == 0) {
+    return(data)
+  }
+  while (length(data) < size) {
+    bytesToRead = size - length(data)
+    extra <- readBin(con, raw(), bytesToRead, endian = "big")
```
davies: endian is not needed.

pkg/inst/worker/daemon.R
```diff
@@ -0,0 +1,70 @@
...
+
+# Read from connection, retrying to read exactly 'size' bytes
+readWithRetry <- function(con, size) {
+  data <- readBin(con, raw(), size, endian = "big")
+  # Since this function is called after
+  # calling `select`, 0 bytes read implies socket is closed
+  if (length(data) == 0) {
+    return(data)
+  }
+  while (length(data) < size) {
```
davies: This could be dead loop, if the socket is closed.

davies:
I will port these changes into #193, thanks!

pkg/inst/worker/daemon.R
shivaram: Its not really hacky as this is the same technique used by the `makeCluster` code in R [1] . However `q(save="no')` also seems to work fine on the linux VM

[1] https://github.com/wch/r-source/blob/R-3-1-branch/src/library/parallel/R/unix/forkCluster.R#L59

pkg/inst/worker/daemon.R
shivaram: Yeah this is a good point. I originally had a `checkWrite` function where if we read 0 bytes, I'd try to write a few bytes to the socket and add `tryCatch` to return false if the write fails.

I was hoping that we'd not run into this as we always write 4 bytes on the stream but yes we should check for a closed socket if we repeatedly get zero bytes

pkg/inst/worker/daemon.R
davies: This depends on the internal behavior of parallel package, it could be changed in future, so I called it `hacky`.

I tried quit() on Mac, it's weird to see a test failure for pipeRDD, did not figure out the reason. so let's keep this. （also mcfork() are internal API).

pkg/inst/worker/daemon.R
davies: we can limited the retry to 10 or 20 times?

shivaram: Yeah we could do that -- Or we could do another `select` loop here with a read call inside it. That way if `select` returns `TRUE` and `read` returns 0 bytes we know the socket is closed.

shivaram: Actually thinking more about it you could probably refactor the code to let the outer select loop handle this. We can just maintain a global buffer of bytes read so far and pass it in to `readIntPort` -- if that buffer comes back with 4 bytes then convert it and clear the buffer.

davies: There is still a chance that `select()` return `TRUE`, but readBin() is interrupted by signal (for example, SIGCHLD), return 0 bytes.

shivaram:
Closing this in favor of #193

PR #210, opened 2015-03-08
Add checks for namespace access operators (:: and :::) in cleanClosure.

hlin09:


pkg/R/utils.R
```diff
@@ -278,6 +278,8 @@ processClosure <- function(node, oldEnv, argNames, newEnv) {
         }
       } else if (nodeChar == "$") {  # Skip the field.
         processClosure(node[[2]], oldEnv, argNames, newEnv)
+      } else if (nodeChar == "::" || nodeChar == ":::") {
```
shivaram: A slightly unrelated question -- but I was wondering what other character we could run into and the only other thing I could think of is `@`. Do we need to handle `@` somehow or does it not matter ?

shivaram:
Thanks @hlin09 for looking into this. Can we also add a test case where `:::` and `::` are used ?

shivaram:
Thanks for the test case and thanks @elbamos for helping us find this bug.

Merging this

PR #211, opened 2015-03-09
[SPARKR-220] Move setGeneric() to a separate file (00-generics.R) to make S4 generic methods easier to follow, and visible to all data types' implementation.

hlin09:
Adds a new file 00-generics.R.
Moves setGeneric() to 00-generics.R.


shivaram:
Thanks @hlin09 -- Couple of questions I had

1. Since we use the `Collate` field in our package description to control the order in which R files are assembled we can just call this `generics.R` and put it at the top of the list ?

2. Should the roxygen docs be next to the generics or the `setMethod` -- The reason I ask this is that the docs might vary a little bit across RDDs, DStreams and DataFrames etc. so I am wondering whats the best way to handle it.

@concretevitamin Any thoughts on this ?

concretevitamin:
Commenting on @shivaram's second point: not sure if there's any technical reason barring this, but putting the docs (especially the usage examples) above the implementations definitely helps developers as well as users to navigate through the APIs more easily. 

If we can do this, we can just treat `00-generics.R` as a placeholder of boilerplate `setGeneric` code that developers can mostly ignore.

hlin09:
@shivaram  and @concretevitamin, I totally agree with you on this after reexamining the roxygen docs. I guess most of them should go with implementation. I can put all of them back, or maybe just leave `@export` and `@rdname` in the `generics.R`.

hlin09:
I have just done some debugging with some questions:
1. If I set `collate` myself in `DESCRIPTION`, the roxygen will complain with `argument 'isCached' missing`. It seems because we have not give default values in RDD's initializer. We can either provide the default initializer, or remove `collate` in `DESCRIPTION` by using `0-generics.R`.
2. Should we have the methods for different data types in the same Rd file? If so, should developer write multiple roxygen docs (in different method implementation) for the same generic? The resulting Rd file might be fairly verbose and confusing.
3. In some of the APIs we have change the parameter name from `rdd` to `x`. Do we need to update their Rd now. I found it when I was testing with roxygen.

shivaram:
Point-wise replies
1. Yeah lets add `isCached` a default value and manually put it in `collate`. I think manually putting it in `collate` is fine, but we can also use `@include` as used in https://github.com/amplab-extras/SparkR-pkg/blob/sparkr-sql/pkg/R/DataFrame.R#L3

2. I think this depends on the types, but it would be good to have them on the same Rd file wherever possible as the operations should be related. Of course we should try to write these documents avoiding duplication. This is my opinion though and it would be good to hear what others think (cc @cafreeman)

3.  Yeah we should just use `x` everywhere for the parameter names. I think the Rd files were just not regenerated after the change. However in the future we are considering not storing Rd files in the repository but just generating them when we do a release / binary build.

hlin09:
Done. Just a note that I use devtools::documents() instead of roxygenize, and the error goes away.

pkg/R/RDD.R
```diff
@@ -839,6 +727,7 @@ setMethod("take",
             resList
           })
 
+setClassUnion("missingOrInteger", c("missing", "integer"))
```
shivaram: Could we also move this to the generics.R file ? I think we use missingOrInteger and missingOrNumeric in a bunch of places, so it'll be good to unify them

shivaram:
Thanks @hlin09 for the update. This is looking pretty good and I just had a small inline comment about `setClassUnion`. 
@concretevitamin -- Any other comments ?

cc @cafreeman as this might affect the DataFrame stuff in the sparkr-sql branch

concretevitamin:
LGTM :+1:

shivaram:
Merging this

PR #212, opened 2015-03-11
[SPARKR-192]  createDataFrame from rdd

davies:
This PR introduce createDataFrame() and toDF() to create DataFrames from existing RDDs.


pkg/R/SQLContext.R
```diff
@@ -1,5 +1,130 @@
 # SQLcontext.R: SQLContext-driven functions
 
+infer_type <- function(x) {
+  if (is.null(x)) {
+    stop("can not infer type from NULL")
+  }
+  type <- switch(class(x),
+                 integer = "long",
```
shivaram: Just curious: does SparkSQL not support integer type ?

pkg/R/SQLContext.R
```diff
@@ -1,5 +1,130 @@
...
+                 character = "string",
+                 logical = "boolean",
+                 double = "double",
+                 numeric = "double",
+                 raw = "binary",
+                 list = "array",
+                 environment = "map",
+                 stop("Unsupported type for DataFrame"))
+
+  if (type == "map") {
```
shivaram: Can you add a short comment about the various null fields we have ? I see `valueContainsNull`, `containsNull`, `nullable` etc.

pkg/R/SQLContext.R
davies: In PySpark and JsonRDD/JsonFile, we changed to use LongType for integer in Python and JSON, because IntegerType is 32bit, but int in Python/JSON is 64bit (in 64-bit machine). It could overflow if we use IntegerType.

davies: Just realized that the integer in R is only 32-bit (on 64bit machine), will change to use IntegerType in SparkR.

pkg/R/SQLContext.R
```diff
@@ -1,5 +1,130 @@
...
+  if (is.null(schema) || is.null(names(schema))) {
+    row <- first(rdd)
+    names <- if (is.null(schema)) {
+      names(row)
+    } else {
+      schema
+    }
+    if (is.null(names)) {
+      names <- lapply(1:length(row), function(x) {
+       paste("_", as.character(x), sep="")
```
shivaram: style nit: one more space before `paste` and `sep = ""` (space before =)

pkg/R/SQLContext.R
```diff
@@ -1,5 +1,130 @@
...
+    fields <- lapply(1:length(row), function(i) {
+      list(name = names[[i]], type = types[[i]], nullable = TRUE)
+    })
+    schema <- list(type = "struct", fields = fields)
+  }
+
+  stopifnot(class(schema) == "list")
+  stopifnot(schema$type == "struct")
+  stopifnot(class(schema$fields) == "list")
+  schemaString <- as.character(jsonlite::toJSON(schema, auto_unbox = TRUE))
```
shivaram: Hmm - are there ways other than json to serialize the schema ? We tend to avoid adding new dependencies to the project if possible and this would require `jsonlite` as a dependency. 
For example: since this is just a named list can we use our existing `writeObject`, `readObject` methods ?

pkg/R/deserialize.R
```diff
@@ -122,21 +123,24 @@ readRowList <- function(obj) {
   # the numCols bytes inside the read function in order to correctly
   # deserialize the row.
   rawObj <- rawConnection(obj, "r+")
-  numCols <- SparkR:::readInt(rawObj)
-  rowOut <- SparkR:::readRow(rawObj, numCols)
-  close(rawObj)
-  rowOut
+  on.exit(close(rawObj))
```
shivaram: A slightly better way to address this would be to pass in `rawObj` as an argument to `readRowList` and close it after the function returns. (This is only used in utils.R)

shivaram: Actually that would be different -- it would try to reuse the `rawConnection`. I am not sure if we can reuse it safely though.

pkg/R/serialize.R
```diff
@@ -55,6 +55,29 @@ writeRawSerialize <- function(outputCon, batch) {
   writeRaw(outputCon, outputSer)
 }
 
+writeRowSerialize <- function(outputCon, rows) {
+  lapply(rows, function(r) {
```
shivaram: you can wrap the `invisible` around the lapply

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/RRDD.scala
```diff
@@ -255,14 +254,15 @@ private class PairwiseRRDD[T: ClassTag](
 private class RRDD[T: ClassTag](
     parent: RDD[T],
     func: Array[Byte],
-    parentSerializedMode: String,
+    deserializer: String,
+    serializeMode: String,
```
shivaram: lets also call this serializer to be consistent ?

shivaram:
Thanks @davies for the change. This is looking great. I left some inline comments and the other thing I wanted to mention is that it'll be good to add some unit tests for this.

pkg/R/deserialize.R
davies: I didn't get it, why it will try to reuse rawConnection? I see `on.exit` as something as `finally {}` in Java.

shivaram: It doesn't reuse it right now. I was wondering if we could reuse it by passing it as an argument. Using `on.exit` as a replacement for `finally` is fine.

shivaram:
Thanks @davies for the update. I'll take a more detailed look tonight, but I was wondering what the recursive serialization bug was ?

@sun-rui Could you also take a look at the SerDe changes here ?

davies:
@shivaram Right now, we do not support to serialize a list of objects with different types, it's needed to support nested structure.

shivaram:
Hmm - is it that we have lists with both boolean and string in the schema ? I was thinking the `toJSON` will convert everything to a string

pkg/R/SQLContext.R
```diff
@@ -1,5 +1,179 @@
...
+#' }
+
+# TODO(davies): support sampling and infer type from NA
+createDataFrame <- function(sqlCtx, data, schema = NULL, samplingRatio = 1.0) {
+  if (is.data.frame(data)) {
+      schema <- names(data)
+      n <- nrow(data)
+      m <- ncol(data)
+      # get rid of factor type
+      dropFactor <- function(x) {
```
shivaram: I think `as.character(x)` should be enough to convert a factor into strings.

pkg/R/SQLContext.R
```diff
@@ -1,5 +1,179 @@
...
+  }
+  if (is.list(data)) {
+    sc <- callJStatic("edu.berkeley.cs.amplab.sparkr.SQLUtils", "getJavaSparkContext", sqlCtx)
+    data <- parallelize(sc, data)
+  }
+  stopifnot(inherits(data, "RDD"))
+
+  rdd <- data
+  if (is.null(schema) || is.null(names(schema))) {
+    row <- first(rdd)
```
shivaram: this is a minor point, but in the case the the data passed in is a data.frame can't we get the schema in the loop before ?  If I follow the code correctly we will only get the column names there and we will still do `first` on the RDD here.

pkg/R/SQLContext.R
davies: `x` could be factor or not, as.character(x) will also turn numeric into character.

pkg/R/SQLContext.R
davies: Yes, we can do it before parallelize(), can do it later, but i think they do not have much difference.

davies:
For the schema part, it support nested structures. But when we serialize the rdd in `row` mode, it does not, for example, a row could be list(1, 'a', list(1, 'a')), then it will fail.

pkg/R/SQLContext.R
shivaram: Yeah I see that -- I meant inside the if clause, instead of `levels(x)[x]` we can do `as.character`

davies: Is it possible that the actual value could be something else? for example, Date or POSIXlt?

pkg/R/SQLContext.R
```diff
@@ -1,5 +1,179 @@
...
+#' sc <- sparkR.init()
+#' sqlCtx <- sparkRSQL.init(sc)
+#' rdd <- lapply(parallelize(sc, 1:10), function(x) list(a=x, b=as.character(x)))
+#' df <- createDataFrame(sqlCtx, rdd)
+#' }
+
+# TODO(davies): support sampling and infer type from NA
+createDataFrame <- function(sqlCtx, data, schema = NULL, samplingRatio = 1.0) {
+  if (is.data.frame(data)) {
+      schema <- names(data)
```
shivaram: The schema variable is not used in this code block and as we discussed below we don't really use this information anyways (as names(schema) will be null ?). I think we can remove it as its confusing if its not used

shivaram:
Hmm -- I was thinking SparkSQL did not support heterogenous lists, so it wouldn't be a problem. Is this to support `StructTypes` instead ? One way to handle StructType is to make it recursive and just call serialize row again ?

pkg/R/SQLContext.R
davies: It will be used later. We did not put the names of column into RDD.

pkg/R/SQLContext.R
shivaram: Yeah but I don't think we can do much if its factor -- and as far as I can see `levels(x)` always creates strings out of the provided type
```
> f <- factor(c(Sys.time(), Sys.time() + 1))
> levels(f)
[1] "2015-03-12 23:22:57" "2015-03-12 23:22:58"
```

pkg/R/SQLContext.R
shivaram: Ah cool - Could you add a comment here ?

davies:
Right now, when we serialize a Row, we just call `writeObject` for each of them without any knowledge of schema.

I think it could be another PR (need some refactor in serializer), also we need to figure out how to support them in data.frame.

pkg/R/SQLContext.R
davies: It's weird, how can we get the original types back?

shivaram:
Hmm okay -- if its complicated lets hold-off on supporting `StructField` types in this PR. I'll take another look at this tomorrow.

pkg/R/SQLContext.R
```diff
@@ -1,5 +1,180 @@
...
+                 POSIXct = "timestamp",
+                 stop(paste("Unsupported type for DataFrame:", class(x))))
+
+  if (type == "map") {
+    stopifnot(length(x) > 0)
+    key <- ls(x)[[1]]
+    list(type = "map",
+         keyType = "string",
+         valueType = infer_type(get(key, x)),
+         valueContainsNull = TRUE)
```
shivaram: so while we set `valueContainsNull = TRUE`, am I right that currently we don't handle NULL or NA values ?

pkg/R/SQLContext.R
```diff
@@ -1,5 +1,180 @@
...
+      paste('{', d, '}', sep = '')
+    } else {
+      l <- paste(lapply(x, tojson), collapse = ', ')
+      paste('[', l, ']', sep = '')
+    }
+  } else if (is.character(x)) {
+    paste('"', x, '"', sep = '')
+  } else if (is.logical(x)) {
+    if (x) "true" else "false"
+  }
```
shivaram: might be good to add a `else` clause here that says `stop("unexpected type in JSON")`

pkg/R/SQLContext.R
shivaram: Factor variables are similar to Enums in Java -- so when you create a factor variable with say levels L1, L2, L3 then the objects L1, L2, L3 are converted to strings to give them as level names. So in some sense there is no type to get back.  
I guess abetter question to ask is what would be a good type matching for Enums in SparkSQL. Right now character seems fine to me.

pkg/R/SQLContext.R
```diff
@@ -1,5 +1,180 @@
...
+          x
+        }
+      }
+      data <- lapply(1:n, function(i) {
+        lapply(1:m, function(j) { dropFactor(data[i,j]) })
+      })
+  }
+  if (is.list(data)) {
+    sc <- callJStatic("edu.berkeley.cs.amplab.sparkr.SQLUtils", "getJavaSparkContext", sqlCtx)
+    data <- parallelize(sc, data)
```
shivaram: minor style thing: instead of reassigning to `data`, I'd prefer something like
 
```
if (is.list(data)) {
  rdd <- parallelize(sc, data)
} else if (inherits(data, "RDD")) {
  rdd <- data
} else {
  stop(paste("cannot create DF from ", class(data), "\n")))
}
```

pkg/R/SQLContext.R
```diff
@@ -1,5 +1,180 @@
...
+  }
+  stopifnot(inherits(data, "RDD"))
+
+  rdd <- data
+  if (is.null(schema) || is.null(names(schema))) {
+    row <- first(rdd)
+    names <- if (is.null(schema)) {
+      names(row)
+    } else {
+      as.list(schema)
```
shivaram: shouldn't this be `names(schema)` in the case that the user passes in a schema object ? (I can see that for the data frame case we will save the names in schema from before)

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SerDe.scala
```diff
@@ -76,6 +78,15 @@ object SerDe {
     val intVal = in.readInt()
     if (intVal == 0) false else true
   }
+  def readDate(in: DataInputStream) = {
```
shivaram: minor nit: new line before this function starts

pkg/src/src/main/scala/edu/berkeley/cs/amplab/sparkr/SerDe.scala
```diff
@@ -233,6 +252,14 @@ object SerDe {
     val intValue = if (value) 1 else 0
     out.writeInt(intValue)
   }
+  def writeDate(out: DataOutputStream, value: Date) {
```
shivaram: newline before this

shivaram:
@davies - Just took another pass. This is looking pretty good. The main comments I had were over what kind of fields we support right now and what we dont. It'll be good to open JIRAs for things we don't support right now. These issues will not be high priority but its better to track them to avoid confusion later.

davies:
created https://sparkr.atlassian.net/browse/SPARKR-230

pkg/R/SQLContext.R
davies: Spark SQL currently does not use this (`valueContainsNull`), it's designed for future.

pkg/R/SQLContext.R
davies: we already check that the `names(schema)` is null.

pkg/R/SQLContext.R
davies: done, thanks for the explain.

davies:
@shivaram Is this ready to go?

pkg/R/deserialize.R
```diff
@@ -26,10 +26,12 @@ readTypedObject <- function(con, type) {
     "b" = readBoolean(con),
     "d" = readDouble(con),
     "r" = readRaw(con),
+    "D" = readDate(con),
```
shivaram: Can you update the type mapping at the top of this file ? Its good to keep code comments up to date. Similar updates in serialize.R and SerDe.scala

pkg/R/serialize.R
```diff
@@ -16,10 +16,11 @@ writeObject <- function(con, object, writeType = TRUE) {
   # NOTE: In R vectors have same type as objects. So we don't support
   # passing in vectors as arrays and instead require arrays to be passed
   # as lists.
+  type <- class(object)[[1]]  # class of POSIXlt is c("POSIXlt", "POSIXt")
   if (writeType) {
-    writeType(con, class(object))
+    writeType(con, type)
   }
-  switch(class(object),
+  switch(type,
```
shivaram: can you add a type for `NULL` ? We'll then need a `readObject` on the Scala side that will parse the `NULL` and then convert it to the null type for SparkSQL.

pkg/R/serialize.R
```diff
@@ -16,10 +16,11 @@ writeObject <- function(con, object, writeType = TRUE) {
...
          integer = writeInt(con, object),
          character = writeString(con, object),
          logical = writeBoolean(con, object),
```
shivaram: So it turns out `class(NA)` returns `logical` in R. So in `writeBoolean` we can add a `is.na` check and write out the same NullType for SparkSQL

shivaram:
Thanks @davies - The updates look good. I just had one last concern with respect to how we handle NA. Basically right now while we go from DF to RDD we convert nulls to NAs. So it'll be great if we can similarly convert NA in R to null in SparkSQL

davies:
NULL is different than NA, currently we only support `null` in SparkSQL.

NA is a bit difficult, we need more thinking on it, could leave it later.

davies:
I don't want this PR block #225 too much, it wastes our time to fix conflicts again and again.

shivaram:
Okay I filed https://sparkr.atlassian.net/browse/SPARKR-232 to track NAs. It turns out `NULL` isn't something you can even set in a dataframe in R. For example
```
> iris[1,1] <- NULL
Error in `[<-.data.frame`(`*tmp*`, 1, 1, value = NULL) :
  replacement has length zero
```
So its more important to handle NAs rather than NULLs, but we can revisit it in a later PR.

shivaram:
Rest of PR LGTM. Merging this

davies:
@shivaram I saw that, NULL is so special in R, it can not be in a collection, for example length(c(1, NULL, 2)) == 2

PR #213, opened 2015-03-12
[SPARKR-224] The serialized flag in a PipelinedRDD is not set correctly.

sun-rui:
This PR fix the issue reported at https://sparkr.atlassian.net/browse/SPARKR-224

### DIFF
--- a/pkg/R/RDD.R
+++ b/pkg/R/RDD.R
@@ -41,13 +41,16 @@ setMethod("initialize", "RDD", function(.Object, jrdd, serialized,
   .Object
 })
 
-setMethod("initialize", "PipelinedRDD", function(.Object, prev, func, jrdd_val) {
+setMethod("initialize", "PipelinedRDD", function(.Object, prev, func) {
   .Object@env <- new.env()
   .Object@env$isCached <- FALSE
   .Object@env$isCheckpointed <- FALSE
-  .Object@env$jrdd_val <- jrdd_val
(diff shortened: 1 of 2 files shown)


shivaram:
@sun-rui So we actually ran into this problem in the sparkr-sql branch and I refactored the code to introduce a `getSerializedMode` function in https://github.com/amplab-extras/SparkR-pkg/commit/71a73b2561d39df045e7517b675b22ac98079586

Couple of points:
- Can you test to see if that fixes the problem you are seeing ? If it works, can you port the change from sparkr-sql to master ?  
- There are a few other changes which might need to be ported from sparkr-sql to master. It'll be great if we open a JIRA to track this. As a part of that lets also do a merge from master to sparkr-sql to get both branches to a similar state (except for the DataFrame methods which will be in that branch).

sun-rui:
@shivaram, I verified that on the sparkr-sql branch, the bug in unionRDD() is fixed. But I am not sure if it is right to make a backport to master now. As davies submitted a JIRA to merge master into sparkr-sql, should I wait for him to complete it and then start back-porting so that we can reduce duplicate effort?

shivaram:
Hmm sure. I don't mind waiting for one side of the merge to finish before doing the other one. I created https://sparkr.atlassian.net/browse/SPARKR-228 to track this

sun-rui:
Close this PR.

PR #214, opened 2015-03-12
[SPARKR-156] Phase 3: Implement zip() in the RDD class.

sun-rui:
zipRDD() is named due to the name confilict with R's built-in zip() function.

There is also an alternative implementation @ https://github.com/amplab-extras/SparkR-pkg/compare/master...sun-rui:zip, which uses join() for zip. I think this one is better.


pkg/R/RDD.R
```diff
@@ -1323,6 +1323,76 @@ setMethod("zipWithIndex",
...
+              stop("Can only zip RDDs which have the same number of partitions.")
+            }
+
+            if (x@env$serialized != other@env$serialized || x@env$serialized) {
+              # Even if any RDD is serialized, we need to reserialize it
+              # to make sure its partitions are encoded as a single byte array.
+              # For example, partitions of an RDD generated from partitionBy() may be
+              # encoded as multiple byte arrays.
+              x <- reserialize(x)
+              other <- reserialize(other)
```
sun-rui: Do you think we re-serialize the result RDD in partitionBy() so that all serialized RDDs have partitions encoded as a single byte array, so that logic here can be eliminated?

shivaram:
@concretevitamin could you also take a look at this ?

pkg/inst/tests/test_rdd.R
```diff
@@ -405,6 +405,25 @@ test_that("pipeRDD() on RDDs", {
   expect_equal(actual, expected)
 })
 
+test_that("zipRDD() on RDDs", {
```
concretevitamin: Can we add a test case for zipping RDDs of different element types?

concretevitamin:
@sun-rui At a quick glance, the current implementation allows zipping RDDs of different sizes. In vanilla Spark, if we do the following
```
val xs = sc.parallelize(1 to 10)
val ys = sc.parallelize(1 to 20)
xs.zip(ys).collect()
```
we would've gotten `org.apache.spark.SparkException: Can only zip RDDs with same number of elements in each partition`. I think we should adhere to this semantics.

pkg/inst/tests/test_rdd.R
sun-rui: Yes, I did want to add such test cases. But they are blocked by the bug reported by SPARKR-224, which has been fixed in the sparkr-sql branch but have not yet been ported back to the master branch. I will add such test cases after porting changes from sparkr-sql to master.

sun-rui:
@concretevitamin , because zipRDD() internall uses spark zip(), so such error check will be done within zip(). No need to do it again on R side.

shivaram:
@sun-rui Thats not quite true as in SparkR we encode the entire partition as one key (for byte serialization). In that case on the Spark side it will only see both RDDs as having one key. So we should add a check in the R code as well.

sun-rui:
@shivaram , yeah, you are right. I will do it.

pkg/R/RDD.R
```diff
@@ -1323,6 +1323,88 @@ setMethod("zipWithIndex",
...
+            if (n1 != n2) {
+              stop("Can only zip RDDs which have the same number of partitions.")
+            }
+
+            # Check if corresponding partitions of both RDDs have the same
+            # number of elements.
+            getPartitionLengthFunc <- function(part) {
+              list(length(part))
+            }
+            partitionLengths1 <- collect(lapplyPartition(x, getPartitionLengthFunc))
```
shivaram: Hmm - so one thing about this approach is that it will trigger a Spark job on both the RDDs. Could we do something in the task like what is done in Scala version ? [1] i.e. inside the map function we can check if the two lengths are equal or exit with `stop(unequal partitions)` ?

[1] https://github.com/apache/spark/blob/c49d156624624a719c0d1262a58933ea3e346963/core/src/main/scala/org/apache/spark/rdd/RDD.scala#L746

sun-rui: If we want to avoid triggering Spark jobs, alternative approach is to add a special processing mode into worker so that a worker is aware of the data being processing is from Spark zip() operation and it will check the two lengths are equal and after that, it performs further zip of keys, values.

@davies, @shivaram , do you agree with this direction?  Also to support cartesian(), another processing mode can be added for it into worker, the worker will generate product of keys,values in this mode.

pkg/R/RDD.R
shivaram: I don't completely understand the need for a different mode. Why can't we do the check in `partitionFunc` ?

pkg/R/RDD.R
shivaram: I don't completely understand the need for a different mode. Why can't we do the check in `partitionFunc` here ?

sun-rui: After calling Spark zip(), for serialized RDDs, the content of each partition is of (Array[Byte], Array[Byte]). And in RRDD, when it is passing a tuple2 into worker, it will break the tuple as two elements and pass them to the worker separately:
{code}
             case (key, value) =>
               writeElem(key)
               writeElem(value)
{code}

In worker, it calls readDeserialize() to deserialized the two tuple elements. readDeserialize() will combine the two lists into one list , that is (K1, K2,...,Kn,V1, V2,...,Vm). The problem is that there is no boundary information between Kn and V1. So in partitionFunc(), we can't know the length of keys, and length of values. We can only assume that keys and values have the same length.

davies: In order to call JavaRDD.zip(), we need to make sure that both `x` and `y` have the same serialization mode and the same number of items in each partitions (you can have multiple Array[Byte] in one partition after coalesce()). So we always need to call `reserialize`.

If we want to avoid to introduce another serialization mode (it's not too bad actually), we need a way to tell which item come from `x` or `y`. so we could do something like this:
```
append_length <- function(items) { items[[length(items) + 1]] <- length(items) + 1; items }
x <- lapplyPartitions(x, append_length)
y <- lapplyPartitions(y, append_length)
jrdd <- callJMethod(getJRDD(x), "zip", getJRDD(y))
r <- RDD(jrdd)
lapplyPartitions(r, function(items) {
  l <- length(items)
  length_y <- items[[l]]
  length_x <- items[[l - length_y]]
  stopifnot(l == length_x + length_y)
  if (length_x != length_y) {
    stop("can not zip")
  }
  lapply(1:(length_x-1), function(i) {
    list(items[[i]], items[[length_x + i]])
  })
```

sun-rui:
@davies , thank you for your idea of appending partition length to each partition. This can avoid change to worker.R. For now, this approach works. We may consider if special mode in worker is needed in the future.

davies:
LGTM. Could you rebase this PR to sparkr-sql branch?

shivaram:
@davies Lets not try and rebase already open PRs to sparkr-sql branch. While merging this PR into master I will also cherry-pick it into the sparkr-sql branch. We do need these PRs in both branches after all

shivaram:
Thank you @sun-rui for the update. The change looks good to me as well. 
Do you want to finish the sparkr-sql to master merge (so that `getSerializationMode` works correctly etc.) before merging this ?

sun-rui:
@shivaram , yes, after that, I can add more test cases to this patch.

shivaram:
@sun-rui Is this good to go or are you working on more test cases. And the merge to master LGTM

sun-rui:
I am adding more test cases.

sun-rui:
@shivaram , I am done, please review and merge it.

shivaram:
Thanks. LGTM. Merging this and cherry picking into sparkr-sql branch

PR #216, opened 2015-03-12
[SPARKR-189] add more selecter for DataFrame

davies:
# Selecting one column
  df[[1]]
  df[["V1"]]

  # Selecting one or multiple columns as a DataFrame
  df[,1]
  df[,1:2]
  df[,c("V1", "V2")]

   # Column assignment, mutating the DF
   df$V3 <- df$V1 + df$V2

### DIFF
--- a/pkg/R/DataFrame.R
+++ b/pkg/R/DataFrame.R
@@ -774,21 +774,85 @@ setMethod("foreachPartition",
           })
 
 
-############################## DSL ##################################
+############################## SELECT ##################################
+
+getColumn <- function(x, c) {
+  column(callJMethod(x@sdf, "col", c))
+}
 
 setMethod("$", signature(x = "DataFrame"),
           function(x, name) {
-            column(callJMethod(x@sdf, "col", name))
+            getColumn(x, name)
           })
 
+setMethod("$<-", signature(x = "DataFrame"),
+          function(x, name, value) {
+            stopifnot(class(value) == "Column")
+            cols <- columns(x)
+            if (name %in% cols) {
+              cols <- lapply(cols, function(c) {
+                if (c == name) {
+                  alias(value, name)
+                } else {
+                  col(c)
+                }
+              })
+              nx <- select(x, cols)
+            } else {
+              nx <- withColumn(x, name, value)
+            }
+            x@sdf <- nx@sdf
+            x
+          })
+
+setMethod("[[", signature(x = "DataFrame"),
+          function(x, i) {
+            if (is.numeric(i)) {
+              cols <- columns(x)
+              i <- cols[[i]]
+            }
+            getColumn(x, i)
+          })
+
+setMethod("[", signature(x = "DataFrame", i = "missing"),
+          function(x, i, j, ...) {
+            if (is.numeric(j)) {
+              cols <- columns(x)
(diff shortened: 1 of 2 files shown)


shivaram:
@davies This is looking great. One minor comment I had was that it would be good to add an example for the `[[` and `[` usages too -- It can be in the same rd file `select`

@cafreeman Could you also take a look at this PR ?

pkg/R/DataFrame.R
```diff
@@ -774,21 +774,85 @@ setMethod("foreachPartition",
...
+  column(callJMethod(x@sdf, "col", c))
+}
 
 setMethod("$", signature(x = "DataFrame"),
           function(x, name) {
-            column(callJMethod(x@sdf, "col", name))
+            getColumn(x, name)
+          })
+
+setMethod("$<-", signature(x = "DataFrame"),
```
cafreeman: This is cool! Very good idea.

pkg/R/DataFrame.R
```diff
@@ -774,21 +774,85 @@ setMethod("foreachPartition",
...
+          function(x, name, value) {
+            stopifnot(class(value) == "Column")
```
cafreeman: Does `value` need to be limited to `Column` exclusively? I know that we'd want to pass a Column expression in the scenario where we're mutating a column, but I could also see people wanting to use this as a shortcut for `alias` in which case it would nice to be able to pass in a string for `value`.

pkg/R/DataFrame.R
```diff
@@ -774,21 +774,85 @@ setMethod("foreachPartition",
...
+            cols <- columns(x)
+            if (name %in% cols) {
+              cols <- lapply(cols, function(c) {
+                if (c == name) {
+                  alias(value, name)
```
cafreeman: Also, it would be **awesome** if we could assign `NULL` to a column name with this operator and have it drop the column from the DataFrame. This would mirror current `data.frame` behavior in R and the ability to easily drop a column of out of a Spark DataFrame would come in very handy.

pkg/R/DataFrame.R
shivaram: One risk with supporting non-column types is that users might expect behavior similar to R data frames. In R if you do 
```
df$age <- "5"
```
It sets the value "5" to all the rows. Since we don't want to support this (changing all the rows might be very expensive ?) I am fine with keeping this only for columns.

cafreeman: Fair point. That works for me, although I do still think the comment below (about being able to easily drop a column) is worth considering. It came up a lot when I was working through the SparkSQL demo.

pkg/R/DataFrame.R
davies: Good point, I'd like to leave this out this PR. SparkSQL currently does not support drop a column, what we can do is that select all other columns except given one.

davies:
@shivaram I'm not familiar with docs in R, could we leave the docs later? (Go through all of them, fill the missing parts).

shivaram:
Alright I filed a new issue for the column deletion issue at https://sparkr.atlassian.net/browse/SPARKR-227 and we can fill in the docs in SPARKR-213

Merging this

PR #217, opened 2015-03-13
Fix cleanClosure() on recursive function calls.

hlin09:
1. Grouped several test cases together to condense the test code. Added test for def-use chain with interleaving function calls, and recursive function calls. 
2. Fixed the issue with recursive function calls, which led to infinite recursion.
3. Replace global `defVars` (holding the local var names) with local one to `cleanClosure()`. The previous version will clear the `defVars` once `processClosure()` calls cleanClosure() on processing functions. (https://github.com/amplab-extras/SparkR-pkg/blob/master/pkg/R/utils.R#L312)

### DIFF
--- a/pkg/R/utils.R
+++ b/pkg/R/utils.R
@@ -238,51 +238,52 @@ convertEnvsToList <- function(keys, vals) {
 # param
 #   node The current AST node in the traversal.
 #   oldEnv The original function environment.
-#   argNames A character vector of parameters of the function. Their values are
-#            passed in as arguments, and not included in the closure.
-#   newEnv A new function environment to store necessary function dependencies.
-processClosure <- function(node, oldEnv, argNames, newEnv) {
+#   defVars An Accumulator of variables names defined in the function's calling environment,
+#           including function argument and local variable names.
+#   checkedFunc An environment of function objects examined during cleanClosure. It can 
+#               be considered as a "name"-to-"list of functions" mapping.
+#   newEnv A new function environment to store necessary function dependencies, an output argument.
+processClosure <- function(node, oldEnv, defVars, checkedFuncs, newEnv) {
   nodeLen <- length(node)
   
   if (nodeLen > 1 && typeof(node) == "language") {
     # Recursive case: current AST node is an internal node, check for its children. 
     if (length(node[[1]]) > 1) {
       for (i in 1:nodeLen) {
-        processClosure(node[[i]], oldEnv, argNames, newEnv)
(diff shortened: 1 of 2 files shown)


pkg/R/utils.R
```diff
@@ -305,11 +305,27 @@ processClosure <- function(node, oldEnv, argNames, newEnv) {
               !(nodeChar %in% getNamespaceExports("SparkR")))) {  # Only include SparkR internals.
           # Set parameter 'inherits' to FALSE since we do not need to search in
           # attached package environments.
-          if (exists(nodeChar, envir = func.env, inherits = FALSE)) {
+          if (tryCatch(exists(nodeChar, envir = func.env, inherits = FALSE),
```
shivaram: what kinds of errors can `exists` throw ? I am just wondering why we need a `tryCatch` here

pkg/R/utils.R
```diff
@@ -305,11 +305,27 @@ processClosure <- function(node, oldEnv, argNames, newEnv) {
...
+                       error = function(e) { FALSE })) {
             obj <- get(nodeChar, envir = func.env, inherits = FALSE)
-            if (is.function(obj)) {  
-              # if the node is a function call, recursively clean its closure.
-              obj <- cleanClosure(obj)
+            if (is.function(obj)) {  # If the node is a function call.
+              funcList <- mget(nodeChar, envir = checkedFuncs, inherits = F, 
+                               ifnotfound = list(list(NULL)))[[1]]
```
shivaram: Hmm - just so I understand correctly if the funcList is `list(list(NULL))` we will go inside the `sapply` but the `identical(func, obj)` will be `FALSE` ?

shivaram:
Thanks @hlin09 - This is a very useful change. I just had a couple of minor questions about the code inline.

pkg/R/utils.R
hlin09: Yes, you are right. I admit that it is less readable, but I was trying to void extra check for errors returned by `get` if there is no such `nodeChar`. So I use `mget` with default value `list(list(NULL))`.

pkg/R/utils.R
hlin09: At least one error I came across was if the `nodeChar` is `""`, the `get` return a error of `invalid argument`. So I thought that should be a `FALSE` in any case of error.

shivaram: Ah yes - I see that passing in a empty string does throw an error. Having a `tryCatch` is fine here.

shivaram:
Thanks @hlin09 - LGTM. I'll merge this into master and cherry-pick it into `sparkr-sql`.

PR #220, opened 2015-03-14
Include utils before SparkR for `head` to work

shivaram:
Before this change calling `head` on a DataFrame would not
work from the sparkR script as utils would be loaded after SparkR
and placed ahead in the search list. This change requires utils to
be loaded before SparkR

To test this you can try something like the following with and without this change
```
./sparkR
>mockLines <- c("{\"name\":\"Michael\"}",
               "{\"name\":\"Andy\", \"age\":30}",
               "{\"name\":\"Justin\", \"age\":19}")
>jsonPath <- tempfile(pattern="sparkr-test", fileext=".tmp")
>writeLines(mockLines, jsonPath)
>df <- jsonFile(sqlCtx, jsonPath)
>head(df)
```

cc @davies


davies:
LGTM

PR #218, opened 2015-03-13
[SPARKR-225] Merge master into sparkr-sql branch

davies:
There is lots of conflicts, fixed manually line by line, hopefully they are merged correctly.

PR #221, opened 2015-03-15
[SPARKR-219] Clear broadcast variables when JVM is shutdown

shivaram:
This fixes the problem where unit tests could not shutdown SparkContext. Also to help users debug such problems associate a JVM id for each JVM reference we hold. The validity of this id is also checked before a JVM reference is passed to the backend.

cc @davies

### DIFF
--- a/pkg/R/broadcast.R
+++ b/pkg/R/broadcast.R
@@ -60,3 +60,10 @@ setBroadcastValue <- function(bcastId, value) {
   bcastIdStr <- as.character(bcastId)
   .broadcastValues[[bcastIdStr]] <- value
 }
+
+#' Helper function to clear the list of broadcast variables we know about
+#' Should be called when the SparkR JVM backend is shutdown
+clearBroadcastVariables <- function() {
+  bcasts <- ls(.broadcastNames)
+  rm(list = bcasts, envir = .broadcastNames)
+}
--- a/pkg/R/jobj.R
+++ b/pkg/R/jobj.R
@@ -8,6 +8,15 @@
 # List of object ids to be removed
 .toRemoveJobjs <- new.env(parent = emptyenv())
 
+# Check if jobj was created with the current SparkContext
+isValidJobj <- function(jobj) {
+  if (exists(".scStartTime", envir = .sparkREnv)) {
+    jobj$appId == get(".scStartTime", envir = .sparkREnv)
+  } else {
+    FALSE
+  }
+}
+
 getJobj <- function(objId) {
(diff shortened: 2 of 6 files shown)


davies:
Should we also check the `jobj` whether it is valid or not before remove (release) them from JVM (just ignore it)?

davies:
Also, should you change this PR to sparkr-sql branch to smooth the merge?

I think we should freeze master branch (or only important bug fix), do the development all on sparkr-sql branch.

shivaram:
1. Regarding the checking on remove, I had it at first but its a harmless on the JVM side (just a hash map remove) and added unnecessary complexity (we need to track the appId in the toRemoveObjs etc.)

2. Lets hold off on this PR. We'll first merge #218 and after that I'll move this to the sparkr-sql branch. BTW I'll probably also rename it to `branch-1.3` after #218 goes in

davies:
It may be harmful in the case that the jobj created on previous JVM has the same id with object on new JVM. for example: 

```
start()
rdd1 <- xxx
stop()
start()
rdd2 <- xxx  # same id with rdd1
rdd1 <- xxx  # release old `rdd1`
...
lapply(rdd2) # rdd2 may become invalid
```

Because we delay the removal of jobj from JVM, it may work fine in simple test (the one we currently have), but failed in more complicated cases.

Could add a sparkR.stop() at the end of each test file, and see what happen?

davies:
The reason that I asked this is the same test run fine but failed later in other tests (a few days ago).

shivaram:
Yeah thats a good point and I think the right way to handle is to (a) clear the `.toRemoveJObjs` in `sparkR.stop` (it should happen when the last RPC goes through, but its good to do anyways) and (b) in the finalizer in `jobj` we should if check the `appId` matches before adding it to `.toRemoveJObjs`. 

I'll add this tomm and try to come up with a test for this as well.

shivaram:
I added a change to clear the outstanding `jobj` removes and also added a check of the `appId` before queuing up new ones. I added a test case which two scenarios, one where the RDD is gc'd before and one after. It turns out both of these didn't actually cause crashes in the existing code base because the counter in `.validJobjs` would prevent the active `jobj` from getting removed.
@davies Let me know if you have another test case in mind.

davies:
LGTM, I had add sparkR.stop() in the end of all the test file, they all run fine. (We will not needed to slow down the unit test process)

shivaram:
Thanks - Merging this into master and cherry-picking to sparkr-sql

PR #222, opened 2015-03-16
[SPARKR-233] [SPARKR-190] add more API for Column

davies:
Also replace "." as "_" in column names


shivaram:
@davies Looks like there is an test failure here in the newly added test case

pkg/NAMESPACE
```diff
@@ -127,10 +127,17 @@ exportClasses("Column")
 
 exportMethods("asc",
               "desc",
-              "first",
               "last",
               "lower",
               "upper",
+              "like",
```
shivaram: Probably not related to this PR, but could you sort this list ?

shivaram:
Thanks for the change. I just had a minor comment, otherwise LGTM.

pkg/NAMESPACE
davies: I'd like to it later, sort all of them.

shivaram: Ok - It'll be great if you could also sort the generics file (sorted per section)

shivaram:
Merging this

PR #224, opened 2015-03-18
[SPARKR-224] The serialized flag in a PipelinedRDD is not set correctly.

sun-rui:
The issue reported in https://sparkr.atlassian.net/browse/SPARKR-224 was fixed by https://sparkr.atlassian.net/browse/SPARKR-228. Here I add a new test case to unionRDD(), which helped to catch the issue.


shivaram:
LGTM. Thanks @sun-rui Merging this and cherry-picking to `sparkr-sql`

PR #227, opened 2015-03-20
Add SparkR API doc to SparkR github page

felixcheung:
knitr generated doc


index.html
```diff
@@ -128,6 +128,11 @@
 <pre><code>./run-tests.sh</code></pre>
 
 <h2>
+<a name="api-sparkr" class"anchor" href="#api-sparkr"><span class="octicon octicon-link"></span></a>SparkR package API documentation</h2>
+
+<a href="rdocs/00Index.html">Documentation for package ‘SparkR’</a>
```
shivaram: Instead of `00Index.html` I actually think `index.html` looks better ? Also I guess we should say this is for Spark 1.2 but we can do that when we add docs for Spark 1.3

rdocs/00Index.html
```diff
@@ -0,0 +1,484 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN">
```
shivaram: Lets name this directory rdocs-1.2 or rdocs-master to correspond to the github branch name

index.html
felixcheung: That's the name automatically chosen by knitr. I generally prefer not to mess with things but agree that index is nicer looking.

Re 1.3 - yes that's my thought.

--- Original Message ---

rdocs/00Index.html
felixcheung: OK, sure. To clarify though, that would be in the url, is that OK?

--- Original Message ---

index.html
shivaram: Ah I actually meant that knitr generates a `index.html` which has two frames. It is already a part of the PR https://github.com/amplab-extras/SparkR-pkg/pull/227/files#diff-39721893066db73eb1158d7561a111c1 as you can see

rdocs/00Index.html
shivaram: Good point - since its a part of the path can we make it rdocs/1.2/  ?

index.html
felixcheung: ah sorry. I couldn't see that in email. updated, great point. I didn't see the index.html

rdocs/00Index.html
felixcheung: done.

shivaram:
Thanks. LGTM. Merging this

PR #228, opened 2015-03-20
Port create-doc script from amplab-extras/spark/R

felixcheung:


.gitignore
```diff
@@ -23,3 +23,45 @@ SparkR-pkg.Rproj
 # Eclipse Meta Files
 .project
 .classpath
+<<<<<<< HEAD
```
shivaram: This file didn't merge correctly. Actually we don't need updates to gitignore here, so you can revert back to the version we had

shivaram:
Thanks @felixcheung - Merging this to master and sparkr-sql

PR #230, opened 2015-03-21
Doc minor fixes

felixcheung:
DESCRIPTION file from R pkg, spacing in index.html


shivaram:
Thanks for the fix. Merging this

shivaram:
@felixcheung One minor thing I noticed -- So the `DESCRIPTION` link at http://amplab-extras.github.io/SparkR-pkg/rdocs/1.2/index.html downloads a file named `DESCRIPTION` rather than showing it in the browser. Is there anyway we can change this ?

felixcheung:
I'm not sure. Likely the correct content type is not set on the github server and the browser can't figure out what type is it. I will check to see if gh-page can be configured.

--- Original Message ---

felixcheung:
I did some research, it didn't seem so. It is possible that adding/changing the file extension for `DESCRIPTION` will help clue in GitHub Pages the correct content-type of the file, but that would mean we would need to hand modify the knitr generated 00Index.html...

shivaram:
Ah I see - Lets leave it this way for now as I think the package description is also available from within R or the repo itself etc. Hand modifying would mean that we need to update it each time we generate new docs and that seems overkill for now.

felixcheung:
Here's the link for future reference: http://stackoverflow.com/questions/15951012/can-mime-types-of-github-pages-files-be-configured
I can't find any mention of it on GitHub Pages documentation.

Anyway, this PR is mostly to address the ugly 404 :) Thanks for merging the fix!

PR #232, opened 2015-03-24
[SPARKR-199] Change takeOrdered, top to fetch one partition at a time

hqzizania:
Change "takeOrdered" and "top" to fetch one partition at a time like "take".

### DIFF
--- a/pkg/inst/sparkR-submit
+++ b/pkg/inst/sparkR-submit
@@ -62,7 +62,7 @@ cat > /tmp/sparkR.profile << EOF
   projecHome <- Sys.getenv("PROJECT_HOME")
   Sys.setenv(NOAWT=1)
   .libPaths(c(paste(projecHome,"/..", sep=""), .libPaths()))
-  require(SparkR)
+  library(SparkR)
   sc <- sparkR.init()
   assign("sc", sc, envir=.GlobalEnv)
   cat("\n Welcome to SparkR!")
--- a/sparkR
+++ b/sparkR
@@ -27,7 +27,7 @@ cat > /tmp/sparkR.profile << EOF
   projecHome <- Sys.getenv("PROJECT_HOME")
   Sys.setenv(NOAWT=1)
   .libPaths(c(paste(projecHome,"/lib", sep=""), .libPaths()))
-  require(SparkR)
+  library(SparkR)
   sc <- sparkR.init(Sys.getenv("MASTER", unset = ""))
   assign("sc", sc, envir=.GlobalEnv)
   cat("\n Welcome to SparkR!")


pkg/R/RDD.R
```diff
@@ -1062,21 +1063,42 @@ takeOrderedElem <- function(x, num, ascending = TRUE) {
...
+    }
+
+    # a JList of byte arrays
+    partitionArr <- callJMethod(jrdd, "collectPartitions", as.list(as.integer(index)))
+    partition <- partitionArr[[1]]
+
+    # elems is capped to have at most `size` elements
+    elems <- convertJListToRList(partition,
+                                 flatten = TRUE,
+                                 logicalUpperBound = NULL,
```
concretevitamin: Is not knowing the order of the elements the reason we don't pass in `num` to `logicalUpperBound`?

pkg/R/RDD.R
```diff
@@ -1062,21 +1063,42 @@ takeOrderedElem <- function(x, num, ascending = TRUE) {
...
+
+  resList <- list()
+  index <- -1
+  jrdd <- getJRDD(newRdd)
+  numPartitions <- numPartitions(newRdd)
+
+  while (TRUE) {
+    index <- index + 1
+
+    if (index >= numPartitions){
```
concretevitamin: style nit: space after closing paren and before bracket

pkg/R/RDD.R
```diff
@@ -1062,21 +1063,42 @@ takeOrderedElem <- function(x, num, ascending = TRUE) {
...
+    partitionArr <- callJMethod(jrdd, "collectPartitions", as.list(as.integer(index)))
+    partition <- partitionArr[[1]]
+
+    # elems is capped to have at most `size` elements
+    elems <- convertJListToRList(partition,
+                                 flatten = TRUE,
+                                 logicalUpperBound = NULL,
+                                 serializedMode = getSerializedMode(newRdd))
+
+    # TODO: Check if this append is O(n^2)?
```
concretevitamin: I see that this TODO is taken from `take()` (no pun intended), but it'd be good to investigate now   that we're at it (and upmerge is pending) :)

concretevitamin:
LGTM at a high level; I left some minor comments. Thanks @hqzizania!

pkg/R/RDD.R
hqzizania: It seems to me that append has a complexity of O(n) from the source code  https://github.com/wch/r-source/blob/trunk/src/library/base/R/append.R, since it has to create a new List using N members of the old List?

hqzizania:
Sorry, I rebased my fork and did something else.... It seems that I have to reopen a PR.

PR #234, opened 2015-03-25
[SPARKR-163] Support sampleByKey()

hqzizania:
Only the "Support sampleByKey()" commit for this PR.

Please ignore the two "modified: pkg/R/RDD.R" commits, which are for [SPARKR-199].
Sorry about that.

### DIFF
--- a/pkg/R/pairRDD.R
+++ b/pkg/R/pairRDD.R
@@ -783,7 +783,7 @@ setMethod("sortByKey",
 #' sc <- sparkR.init()
 #' rdd <- parallelize(sc, 1:2000)
 #' makePairs <- lapply(rdd, function(x) { if(x%%2 == 0) list("a", x) else list("b", x) })
-#' fractions <- list(c("a", 0.2), c("b", 0.1))
+#' fractions <- list(a = 0.2, b = 0.1)
 #' sample <- sampleByKey(makePairs, FALSE, fractions, 1618L)
 #' 100 < length(lookup(sample, "a")) && 300 > length(lookup(sample, "a")) # TRUE
 #' 50 < length(lookup(sample, "b")) && 150 > length(lookup(sample, "b")) # TRUE
@@ -798,13 +798,6 @@ setMethod("sampleByKey",
           signature(x = "RDD", withReplacement = "logical",
                     fractions = "vector", seed = "integer"),
           function(x, withReplacement, fractions, seed) {
-            funk <- function(k) {
-              k[[1]]
-            }
-
-            funv <- function(v) {
-              v[[2]]
-            }
 
             # The sampler: takes a partition and returns its sampled version.
             samplingFunc <- function(split, part) {
@@ -812,16 +805,18 @@ setMethod("sampleByKey",
               res <- vector("list", length(part))
               len <- 0
 
-              fractionskey <- lapply(fractions, funk)
-              fractionsval <- lapply(fractions, funv)
+              for (elem in fractions) {
+                if (elem < 0.0)
+                  stop(paste("Negative fraction value ", fractions[which(fractions == elem)]))
+              }
 
               # Discards some random values to ensure each partition has a
               # different random seed.
               runif(split)
 
               for (elem in part) {
-                if(elem[[1]] %in% fractionskey){
-                  frac <- as.numeric(fractionsval[which(fractionskey == elem[[1]])])
+                if (elem[[1]] %in% names(fractions)) {
+                  frac <- as.numeric(fractions[which(elem[[1]] == names(fractions))])
                   if (withReplacement) {
                     count <- rpois(1, frac)
                     if (count > 0) {
--- a/pkg/R/pairRDD.R
+++ b/pkg/R/pairRDD.R
@@ -769,4 +769,84 @@ setMethod("sortByKey",
             newRDD <- partitionBy(x, numPartitions, rangePartitionFunc)
             lapplyPartition(newRDD, partitionFunc)
           })
-          
+
+#' @description
+#' \code{sampleByKey} return a subset RDD of the given RDD sampled by key 
+#'
+#' @param x The RDD to sample elements by key, where each element is
+#'             list(K, V) or c(K, V). 
+#' @param withReplacement Sampling with replacement or not
+#' @param fraction The (rough) sample target fraction
+#' @param seed Randomness seed value
+#' @examples
+#'\dontrun{
+#' sc <- sparkR.init()
+#' rdd <- parallelize(sc, 1:2000)
+#' makePairs <- lapply(rdd, function(x) { if(x%%2 == 0) list("a", x) else list("b", x) })
+#' fractions <- list(c("a", 0.2), c("b", 0.1))
+#' sample <- sampleByKey(makePairs, FALSE, fractions, 1618L)
+#' 100 < length(lookup(sample, "a")) && 300 > length(lookup(sample, "a")) # TRUE
+#' 50 < length(lookup(sample, "b")) && 150 > length(lookup(sample, "b")) # TRUE
+#' lookup(sample, "a")[which.min(lookup(sample, "a"))] >= 0 # TRUE
+#' lookup(sample, "a")[which.max(lookup(sample, "a"))] <= 2000 # TRUE
+#' lookup(sample, "b")[which.min(lookup(sample, "b"))] >= 0 # TRUE
+#' lookup(sample, "b")[which.max(lookup(sample, "b"))] <= 2000 # TRUE
+#'}
+#' @rdname sampleByKey
+#' @aliases sampleByKey,RDD-method
+setMethod("sampleByKey",
+          signature(x = "RDD", withReplacement = "logical",
+                    fractions = "vector", seed = "integer"),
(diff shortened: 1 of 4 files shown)


concretevitamin:
You could only include the relevant commit by doing something like
```
# on assist branch
git checkout -b assist-fix
git checkout -
git reset --hard head~3
git cherry-pick assist-fix
git push -f 
```
This should update this PR automatically with the single relevant commit.

pkg/R/pairRDD.R
```diff
@@ -769,4 +769,84 @@ setMethod("sortByKey",
...
+#'             list(K, V) or c(K, V). 
+#' @param withReplacement Sampling with replacement or not
+#' @param fraction The (rough) sample target fraction
+#' @param seed Randomness seed value
+#' @examples
+#'\dontrun{
+#' sc <- sparkR.init()
+#' rdd <- parallelize(sc, 1:2000)
+#' makePairs <- lapply(rdd, function(x) { if(x%%2 == 0) list("a", x) else list("b", x) })
+#' fractions <- list(c("a", 0.2), c("b", 0.1))
```
concretevitamin: What do you think about this interface change: let `fractions` just be a named list. E.g.:
```
fractions <- list(a = 0.2, b = 0.1)
```
This should be easier to use for users.

pkg/R/pairRDD.R
```diff
@@ -769,4 +769,84 @@ setMethod("sortByKey",
...
+              k[[1]]
+            }
+
+            funv <- function(v) {
+              v[[2]]
+            }
+
+            # The sampler: takes a partition and returns its sampled version.
+            samplingFunc <- function(split, part) {
+              set.seed(seed)
```
concretevitamin: It seems to me PySpark is doing [something](https://github.com/apache/spark/blob/master/python/pyspark/rddsampler.py#L31) more sophisticated now

pkg/R/pairRDD.R
```diff
@@ -769,4 +769,84 @@ setMethod("sortByKey",
...
+
+              fractionskey <- lapply(fractions, funk)
+              fractionsval <- lapply(fractions, funv)
+
+              # Discards some random values to ensure each partition has a
+              # different random seed.
+              runif(split)
+
+              for (elem in part) {
+                if(elem[[1]] %in% fractionskey){
```
concretevitamin: nit: spaces

pkg/R/pairRDD.R
```diff
@@ -769,4 +769,84 @@ setMethod("sortByKey",
...
+#' lookup(sample, "a")[which.max(lookup(sample, "a"))] <= 2000 # TRUE
+#' lookup(sample, "b")[which.min(lookup(sample, "b"))] >= 0 # TRUE
+#' lookup(sample, "b")[which.max(lookup(sample, "b"))] <= 2000 # TRUE
+#'}
+#' @rdname sampleByKey
+#' @aliases sampleByKey,RDD-method
+setMethod("sampleByKey",
+          signature(x = "RDD", withReplacement = "logical",
+                    fractions = "vector", seed = "integer"),
+          function(x, withReplacement, fractions, seed) {
```
concretevitamin: We should do a negative fraction check at the beginning here. E.g.: https://github.com/apache/spark/blob/master/python/pyspark/rdd.py#L1859

concretevitamin:
One semantic behavior that we should discuss: with the current patch, if `fractions` omit some keys that actually exist in the RDD, the whole sampling still goes through and just ignores those items:
```
rdd <- parallelize(sc, 1:2000)
makePairs <- lapply(rdd, function(x) { if(x%%3 == 0) list("a", x) else { if (x %% 3 == 1) list("b", x) else list("c", x) }})
fractions <- list(c("a", 0.2), c("b", 0.1)) # omits key "c"
collect(sampleByKey(makePairs, FALSE, fractions, 1618L)) # still okay
```
However, in PySpark the above will throw an error:
```
  File "/Users/zongheng/workspace/spark/python/pyspark/shuffle.py", line 252, in mergeValues
    for k, v in iterator:
  File "/Users/zongheng/workspace/spark/python/pyspark/rddsampler.py", line 118, in func
    if self.getUniformSample() < self._fractions[key]:
KeyError: 'c'

        at org.apache.spark.api.python.PythonRDD$$anon$1.read(PythonRDD.scala:135)
        at org.apache.spark.api.python.PythonRDD$$anon$1.<init>(PythonRDD.scala:176)
        at org.apache.spark.api.python.PythonRDD.compute(PythonRDD.scala:94)
        at org.apache.spark.rdd.RDD.computeOrReadCheckpoint(RDD.scala:277)
        at org.apache.spark.rdd.RDD.iterator(RDD.scala:244)
        at org.apache.spark.api.python.PairwiseRDD.compute(PythonRDD.scala:307)
```

I think we should be consistent with PySpark. 

Also, @hqzizania could you investigate the symmetric case where `fractions` contains a key that does not exist in the RDD, and compare SparkR & PySpark's behaviors?

pkg/R/pairRDD.R
```diff
@@ -769,4 +769,84 @@ setMethod("sortByKey",
...
+              len <- 0
+
+              fractionskey <- lapply(fractions, funk)
+              fractionsval <- lapply(fractions, funv)
+
+              # Discards some random values to ensure each partition has a
+              # different random seed.
+              runif(split)
+
+              for (elem in part) {
```
concretevitamin: I don't see an obvious way to reuse code w/ `sampleRDD`, but there is a lot of duplication here so it'd be good to do so - @hqzizania do you have any thoughts on this?

pkg/R/pairRDD.R
hqzizania: I think you are right and will change it into a named list.

hqzizania:
@concretevitamin I have investigated it. In the case where fractions contains a key that does not exist in the RDD, SparkR (in this current patch) & PySpark (spark-1.3.0) both ignore it and finish the sampling of the keys that existing in the RDD. If want to be consistent with PySpark, the only thing we should improve is to make SparkR throw an error in the case you said like PySpark?

concretevitamin:
@hqzizania Thanks for investigating, indeed that'd be good to do (and we should document these two special behaviors in the method's doc).

pkg/R/pairRDD.R
concretevitamin: @hqzizania just want to make sure you see this comment about XOR

pkg/inst/tests/test_rdd.R
```diff
@@ -620,3 +620,29 @@ test_that("collectAsMap() on a pairwise RDD", {
   vals <- collectAsMap(rdd)
   expect_equal(vals, list(`1` = "a", `2` = "b"))
 })
+
+test_that("sampleByKey() on pairwise RDDs", {
+  rdd <- parallelize(sc, 1:2000) 
+  pairsRDD <- lapply(rdd, function(x) { if(x%%2 == 0) list("a", x) else list("b", x) })
```
concretevitamin: nit: space after if

pkg/R/pairRDD.R
```diff
@@ -769,4 +769,79 @@ setMethod("sortByKey",
...
+#' \code{sampleByKey} return a subset RDD of the given RDD sampled by key 
+#'
+#' @param x The RDD to sample elements by key, where each element is
+#'             list(K, V) or c(K, V). 
+#' @param withReplacement Sampling with replacement or not
+#' @param fraction The (rough) sample target fraction
+#' @param seed Randomness seed value
+#' @examples
+#'\dontrun{
+#' sc <- sparkR.init()
```
concretevitamin: Fixing space/naming nits:
```
+#' sc <- sparkR.init()
+#' rdd <- parallelize(sc, 1:2000)
+#' pairs <- lapply(rdd, function(x) { if (x%%2 == 0) list("a", x) else list("b", x) })
+#' fractions <- list(a = 0.2, b = 0.1)
+#' sample <- collect(sampleByKey(pairs, FALSE, fractions, 1618L))
+#' 100 < length(lookup(sample, "a")) && 300 > length(lookup(sample, "a")) # TRUE
+#' 50 < length(lookup(sample, "b")) && 150 > length(lookup(sample, "b")) # TRUE
+#' lookup(sample, "a")[which.min(lookup(sample, "a"))] >= 0 # TRUE
+#' lookup(sample, "a")[which.max(lookup(sample, "a"))] <= 2000 # TRUE
+#' lookup(sample, "b")[which.min(lookup(sample, "b"))] >= 0 # TRUE
+#' lookup(sample, "b")[which.max(lookup(sample, "b"))] <= 2000 # TRUE
```

pkg/R/pairRDD.R
hqzizania: Yeah, there is a rddsampler.py in pyspark. The RDDSamplerBase in rddsampler.py is reused for a couple of times. But I haven't thought of a nice plan for this sparkR patch. @concretevitamin

concretevitamin:
Thanks. I open https://github.com/hqzizania/SparkR-pkg/pull/2 fixing some style nits; with that this is good to go :)

concretevitamin:
Merging in master & cherry-picking to sparkr-sql.

PR #237, opened 2015-04-01
[SPARKR-154] Phase 2: implement cartesian().

sun-rui:


pkg/R/RDD.R
```diff
@@ -1487,3 +1487,87 @@ setMethod("zipRDD",
...
+                  getSerializedMode(x) == "byte") {
+              # Append the number of elements in each partition to that partition so that we can later
+              # know the boundary of elements from x and other.
+              #
+              # Note that this appending also serves the purpose of reserialization, because even if 
+              # any RDD is serialized, we need to reserialize it to make sure its partitions are encoded
+              # as a single byte array. For example, partitions of an RDD generated from partitionBy()
+              # may be encoded as multiple byte arrays.          
+              appendLength <- function(part) {
+                part[[length(part) + 1]] <- length(part) + 1
```
concretevitamin: This line runs the risk of confusion; maybe this is what we want?
```
len <- length(part)
part[[len + 1]] <- len + 1
```

pkg/R/RDD.R
```diff
@@ -1487,3 +1487,87 @@ setMethod("zipRDD",
...
+                  } else {
+                    values <- list()
+                  }
+                  
+                  mergeCompactLists(keys, values)
+                } else {
+                  keys <- part[c(TRUE, FALSE)]
+                  values <- part[c(FALSE, TRUE)]
+                  mapply(
+                    function(k, v) {
```
concretevitamin: nit: can put this function in one line

pkg/inst/tests/test_rdd.R
```diff
@@ -446,6 +446,49 @@ test_that("zipRDD() on RDDs", {
...
+               list(
+                 list(0, "Spark is pretty."),
+                 list(0, "Spark is awesome."),
+                 list(1, "Spark is pretty."),
+                 list(1, "Spark is awesome.")))
+  
+  rdd1 <- map(rdd, function(x) { x })
+  actual <- collect(cartesian(rdd, rdd1))
+  expect_equal(sortKeyValueList(actual),
+               list(
```
concretevitamin: nit: we could write this list once above

pkg/inst/tests/test_rdd.R
```diff
@@ -446,6 +446,49 @@ test_that("zipRDD() on RDDs", {
   unlink(fileName)
 })
 
+test_that("cartesian() on RDDs", {
```
concretevitamin: Would be good to add a case where one side is empty.

concretevitamin:
Thanks @sun-rui, I think this overall looks good. 

Would be good if others who're familiar with serialization modes can also take a look. /cc @shivaram @cafreeman

pkg/R/RDD.R
sun-rui: fixed. The same logic in zipRDD() is also fixed.

pkg/R/RDD.R
sun-rui: fixed. The same logic in zipRDD() is also fixed.

pkg/inst/tests/test_rdd.R
sun-rui: done

concretevitamin:
Latest changes look good - I will take another look at serialization-related changes if others don't have free cycles; otherwise this should be good to go in.

pkg/R/RDD.R
```diff
@@ -1473,17 +1474,98 @@ setMethod("zipRDD",
...
+#' sortByKey(cartesian(rdd, rdd)) 
+#' # list(list(1, 1), list(1, 2), list(2, 1), list(2, 2))
+#'}
+#' @rdname cartesian
+#' @aliases cartesian,RDD,RDD-method
+setMethod("cartesian",
+          signature(x = "RDD", other = "RDD"),
+          function(x, other) {
+            if (getSerializedMode(x) != getSerializedMode(other) || 
+                  getSerializedMode(x) == "byte") {
```
shivaram: minor style nit - the `getSerializedMode` here should align with the line above

pkg/R/RDD.R
```diff
@@ -1473,17 +1474,98 @@ setMethod("zipRDD",
...
+                part[[len + 1]] <- len + 1
+                part
+              }
+              x <- lapplyPartition(x, appendLength)
+              other <- lapplyPartition(other, appendLength)
+            }
+            
+            cartesianJRDD <- callJMethod(getJRDD(x), "cartesian", getJRDD(other))
+            # The cartesianJRDD's elements are of scala Tuple2 type. The serialized
+            # flag Here is used for the elements inside the tuples.
```
shivaram: nit: Here -> here

shivaram:
Sorry for the delay -- The serialization changes look good. One minor comment I had was that it might be good to refactor zipRDD and cartesian to reuse the same code by pulling the common parts to a function in utils.R

pkg/R/RDD.R
sun-rui: fixed

pkg/R/RDD.R
sun-rui: fixed

shivaram:
@sun-rui Can you bring this up to date with master branch ?

@concretevitamin Will be great if you take another look before merging.

pkg/R/utils.R
```diff
@@ -419,3 +419,82 @@ cleanClosure <- function(func, checkedFuncs = new.env()) {
...
+
+# Common function for implmentation of zipRDD() and cartesian().
+# param
+#   x An RDD.
+#   other An RDD.
+#   operation The specified operation, "zip" or "cartesian".
+#
+# return value
+#   The result RDD.
+doZipOrCartesian <- function(x, other, operation = c("zip", "cartesian")) {
```
shivaram: I'm wondering if there is a better way to do this than have this one large function. i.e can we break this up into say `appendLength` and the second one being `mergePartitions(checkEqualLength, compactResults)`. Then `zip` can look like
```
zipRDDs <- function() {
  appendLengths()
  <call JRDD zip here>
  mergePartitions(checkLengths = T, compactResults = F)
}
```

pkg/inst/tests/test_rdd.R
shivaram: +1 This would be good to have

pkg/R/utils.R
```diff
@@ -419,3 +419,82 @@ cleanClosure <- function(func, checkedFuncs = new.env()) {
...
+          values <- part[(lengthOfKeys + 1) : (len - 1)]                    
+        } else {
+          values <- list()
+        }
+        
+        if (operation == "cartesian") {
+          return(mergeCompactLists(keys, values))
+        }
+      } else {
+        keys <- part[c(TRUE, FALSE)]
```
shivaram: Just checking -- does the `cartesian` code path never come here ? Or do we just not need the `mergeCompactLists` here ?

sun-rui: cartesian code path will go here if the two input RDDs are both text file. If any of RDDs is in byte serializer mode, it will go to call mergeCompactLists().

pkg/R/utils.R
sun-rui: I did think about split the single common function into two common functions as mentioned above. But I felt the logic of the two common functions are related and the smell does not feel so good. Anyway, I put a WIP at https://github.com/amplab-extras/SparkR-pkg/commit/02363086977cbbbd2c159c81dac5d2e9617263b4 for this refactoring for your review. If you prefer this way, I will make the change here.

shivaram: Actually I like the WIP version at https://github.com/amplab-extras/SparkR-pkg/commit/02363086977cbbbd2c159c81dac5d2e9617263b4 as it gives some insight when you read `zip` as to what is going on

@concretevitamin any thoughts on this ?

concretevitamin: I think I also preferred https://github.com/amplab-extras/SparkR-pkg/commit/02363086977cbbbd2c159c81dac5d2e9617263b4 as a single longer function is hard to read.

A general comment is that I think it's good to follow @shivaram's purposed interface `mergePartitions(checkLengths = T, compactResults = F)` (something like this; doesn't have to be exact) instead of `mergePartitions(zip)`. The former focuses on the behaviors and is possible to be reused outside of zip/cartesian?

sun-rui: For now, mergePartitions() needs only parameter. If we define two parameters for it: checkLengths , compactResults, they actually need to be mutually exclusive. If they are provided both true or false, undefined result will be got from the function. I don't know about any reuse of this function for now, so I think one parameter is enough. Until we find that this function can be reused, we can change the parameter definitions. What do you think?

shivaram: Yeah having one parameter is fine for now. Ideally we'd call it `mergeStrategy` which could be `zip` or `merge` etc. but I'm wondering if that is complicating things too much. I'm fine with calling the variable a boolean `zip` for now.

concretevitamin: @sun-rui - that makes sense.

shivaram:
Thanks @sun-rui -- LGTM. @concretevitamin you can take a final look and merge !

pkg/inst/tests/test_rdd.R
```diff
@@ -451,6 +451,49 @@ test_that("zipRDD() on RDDs", {
...
+  # test case where one RDD is empty
+  emptyRdd <- parallelize(sc, list())
+  actual <- collect(cartesian(rdd, emptyRdd))
+  expect_equal(actual, list())
+
+  mockFile = c("Spark is pretty.", "Spark is awesome.")
+  fileName <- tempfile(pattern="spark-test", fileext=".tmp")
+  writeLines(mockFile, fileName)
+  
+  rdd <- textFile(sc, fileName)
```
concretevitamin: To avoid flakiness we should use `textFile(sc, fileName, 1)`?

See https://github.com/apache/spark/pull/5460/

davies: It's need only when we want to saveAsXXX()(because the filenames is not sorted).

It's flaky if you have only one file (this case).

davies:
LGTM, merging this into master and sparkr-sql.

PR #241, opened 2015-04-03
 [SPARKR-92] Phase 1: implement mean(rdd), stdev(rdd), variance(rdd)

hqzizania:


concretevitamin:
I think the current approach is a little bit hard to maintain. Should we use some lightweight OOP for StatCounter? That way `mergeStats()` would be easier to understand and easily extensible (we may also need `sampleStdev()` and `sampleVariance()` etc.).

We could move the design discussion to JIRA.

hqzizania:
The functions "sampleStdev" and "sampleVariance" also could be implemented with only a few lines codes via reusing mergeStats() in this patch. Moreover, the "histogram" function maybe require a lot of additional codes like in pyspark. Thus, I suppose a OOP for StatCounter maybe not very necessary at this stage.

hafen:
If you want a more generic `mergeStats()` under the hood, you could follow as outlined in this paper: http://janinebennett.org/index_files/ParallelStatisticsAlgorithms.pdf, which provides a numerically stable algorithm for calculating any order moment.  You could expose `sampleMean()`, `sampleVariance()`, even `sampleKurtosis()`, etc. to the user and call the same common `mergeStats()`.

pkg/R/utils.R
```diff
@@ -419,3 +419,21 @@ cleanClosure <- function(func, checkedFuncs = new.env()) {
   }
   func
 }
+
+# Merge another StatCounter into this one, adding up the internal statistics.
```
shivaram: We could use some comments here as to what structure `x` and `y` have (i.e. what is x[1] and x[2]).  Also a light-weight method to get OOP is to use named lists like `list(sum=x1, sumsq=x2)` etc.

pkg/R/RDD.R
```diff
@@ -638,6 +638,66 @@ setMethod("minimum",
...
+#' @param x The RDD to get the variance of the values of elements from
+#' @examples
+#'\dontrun{
+#' sc <- sparkR.init()
+#' rdd <- parallelize(sc, 1:3)
+#' variance(rdd) # 0.6666667
+#'}
+#' @rdname variance 
+#' @aliases variance,RDD
+setMethod("variance",
```
shivaram: Any thoughts on why we shouldn't override existing R functions like `mean`, `sd`, `var` ?

pkg/R/utils.R
concretevitamin: +1 on named list

pkg/R/RDD.R
hqzizania: I don't know how the existing R functions are implemented. It seems to be very different between original base functions and these functions because of the Reduce used here.

shivaram:
@hqzizania Would it be possible to re-open this PR against the main Spark repo now ? I did plan to do a automatic move for already open PRs but I feel we have some more things to discuss for this one (like @davies suggestion of using DataFrames). I've already created a JIRA on the Spark side at 
https://issues.apache.org/jira/browse/SPARK-6841

Let me know if this sounds good.

pkg/R/RDD.R
cafreeman: The signatures for the generics are all compatible, I believe. At that point, the implementation won't matter with regard to overriding. As long as the generic knows which version to dispatch on, the rest should be fine.

hqzizania: Thanks, I see. @cafreeman

hqzizania:
@shivaram I've opened it against the main Spark repo. But the fixes, improvements and replies to suggestions could be delayed because I have no time at this moment.

shivaram:
@hqzizania - Thanks for the heads up. We can first discuss the design choices and then come to the implementation. I'm going to close the PR here (we can backport it later though I guess its not required as this is a new feature rather than a bug fix)

PR #242, opened 2015-04-03
[SPARKR-92] Phase 2: implement sum(rdd)

hqzizania:


concretevitamin:
LGTM, merging in master and cherry-picking into `sparkr-sql` branch.

@hqzizania Can you also open a PR against https://github.com/amplab-extras/spark for the same changes as well? This is for upmerging into the mainline Spark repo.

PR #243, opened 2015-04-03
[SPARKR-199] Change takeOrdered, top to fetch one partition at a time 

hqzizania:
Reopen because of my fault, Sorry.

### DIFF
--- a/pkg/R/RDD.R
+++ b/pkg/R/RDD.R
@@ -1075,6 +1075,7 @@ takeOrderedElem <- function(x, num, ascending = TRUE) {
   index <- -1
   jrdd <- getJRDD(newRdd)
   numPartitions <- numPartitions(newRdd)
+  serializedModeRDD <- getSerializedMode(newRdd)
 
   while (TRUE) {
     index <- index + 1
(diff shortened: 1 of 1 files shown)


hqzizania:
About the TODO: Check if this append is O(n^2)?

It seems to me that append has a complexity of O(n) from the source code https://github.com/wch/r-source/blob/trunk/src/library/base/R/append.R, since it has to create a new List using N members of the old List? @concretevitamin

pkg/R/RDD.R
```diff
@@ -1062,21 +1063,42 @@ takeOrderedElem <- function(x, num, ascending = TRUE) {
...
+
+    # a JList of byte arrays
+    partitionArr <- callJMethod(jrdd, "collectPartitions", as.list(as.integer(index)))
+    partition <- partitionArr[[1]]
+
+    # elems is capped to have at most `num` elements
+    elems <- convertJListToRList(partition,
+                                 flatten = TRUE,
+                                 logicalUpperBound = num,
+                                 serializedMode = getSerializedMode(newRdd))
```
concretevitamin: minor: we can just call `getSerializedMode(newRdd)` once outside the loop

concretevitamin:
Thanks, looks like a simple cons to me so not a big deal. I left a minor comment; could you also remove this TODO as well as the one in `take()`?

hqzizania:
The TODO in take() has been removed. Moreover, calling getSerializedMode(x) is in a loop in take(), thus I also fix it as well.

concretevitamin:
Thanks! Merging in master & sparkr-sql branch.

Would be good if you could open a PR against [amplab-extras/spark](https://github.com/amplab-extras/spark/) when you have some cycles.

PR #246, opened 2015-04-09
Fixes combineByKey

hlin09:


shivaram:
Thanks @hlin09. Could you add a unit test with a key type (anything other than int I guess) that would have failed before ?

hlin09:
Test added.

shivaram:
Thanks. Merging this into master and sparkr-sql branches

PR #247, opened 2015-04-09
[SPARKR-244] Fix test failure after integration of subtract() and subtractByKey() for RDD.

sun-rui:
Rename subtract() method of Data Frame to except, which is consistent with DF scala API.


pkg/NAMESPACE
```diff
@@ -122,7 +122,7 @@ exportMethods("columns",
               "show",
               "showDF",
               "sortDF",
-              "subtract",
```
concretevitamin: Minor: move this to its sorted position?

shivaram:
Thanks @sun-rui - @concretevitamin I'll fix the sort order during the merge

PR #254, opened 2016-06-10
Fix a wget option error

luborliu:
It should be a mistake when it comes to wget command. 
wget command does not support progress-bar option, instead, it supports progress option. 
Before the change users will see:

```
wget: unrecognized option '--progress-bar'
Usage: wget [OPTION]... [URL]...
```

So I change it to "wget --progress=bar".


shivaram:
LGTM. Thanks @luborliu -- One thing to note is that this repository is no longer actively maintained and that SparkR is now a part of the mainline Apache project.

